PR #52, opened 2014-11-25
grovel: support files with multiple dots in names

orivej:
This fixes Quicklisp issue [#758](https://github.com/quicklisp/quicklisp-projects/issues/758) with [libssh2.asd](https://bitbucket.org/alxchk/cl-libssh2/src/a040f346354c65bfe218d054c668b948817fa325/libssh2.asd?at=default) that contains `libssh2-libc.cffi.lisp`, which causes error mentioned in the Quicklisp issue: [failtail.txt](http://report.quicklisp.org/cl-libssh2/2014-09-26/failtail.txt).

Without this patch `(namestring (cffi-grovel::exe-filename #P"a.b.c"))` produces an error under SBCL.

### DIFF
--- a/grovel/grovel.lisp
+++ b/grovel/grovel.lisp
@@ -223,9 +223,19 @@ int main(int argc, char**argv) {
 
 (defparameter *exe-extension* #-windows nil #+windows "exe")
 
+(defun change-pathname-type (path type)
+  (if type
+      (make-pathname :type type :defaults path)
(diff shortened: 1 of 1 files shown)


sionescu:
Either we just recommend people to avoid multiple dots in filenames or use UIOP to parse namestrings. Either way this PR is not a good idea.

Close #53

PR #53, opened 2014-11-25
grovel: support files with multiple dots in names (v2)

orivej:
(Alternative patch for #52.)


sionescu:
Closed. See #52 for details.

PR #59, opened 2015-03-07
doc: update example code in the manual

taptap:
In order to restore libcurl example to working state again.


luismbo:
Pushed as 7033913, thanks!

PR #60, opened 2015-03-13
ecl: add %close-foreign-library call implementation.

dkochmanski:
Valid since ECL 15.2.21 upwards. Fixes loading iolib and possibly other
libraries.


luismbo:
Is this change backwards compatible? If not, can we do this in a way that is?

dkochmanski:
Now it should be backwards compatible with additional restart for ignoring error. Library will stay opened then, but it's not critical, since ECL won't open twice the same library, just will increase refcount.

luismbo:
Pushed as bb0d546 (made a minor tweak: s/FFI/CFFI/). Thanks!

PR #62, opened 2015-07-09
ecl: fix second value of %invoke

dkochmanski:
It is needed if user calls invoke for output (ie pkg-config)


sionescu:
Thanks for the patch.

PR #64, opened 2015-07-13
allocate correct type in e-t-f-d-i for built-in types

3b:
i think this is more correct, probably only matters on 32 bit platforms?


sionescu:
Merged.

PR #65, opened 2015-07-29
Add support for C99 _Bool

sionescu:
Hard-coding the size is one more hack that will work on i386/x86_64 but I'm not sure about other architectures. Better support for ABIs is necessary, which would probably mean that the backends would stop deferring to the host for standard type mappings.
At least there's the test suite.

### DIFF
--- a/doc/cffi-manual.texinfo
+++ b/doc/cffi-manual.texinfo
@@ -1940,6 +1940,11 @@ default.
 (convert-from-foreign 1 :boolean) @result{} t
 @end lisp
 
+@ForeignType{:bool}
+
+The @code{:bool} type represents the C99 @code{_Bool} or C++
+@code{bool}. Its size is usually 1 byte except on OSX where it's an @code{int}.
+
 @ForeignType{:wrapper base-type &key to-c from-c}
 
 The @code{:wrapper} type stores two symbols passed to the @var{to-c}
--- a/tests/misc-types.lisp
+++ b/tests/misc-types.lisp
@@ -86,6 +86,43 @@
           (bool-xor nil nil))
   (t t t t nil t nil))
 
+(defcfun "sizeof_bool" :unsigned-int)
+
+(deftest misc-types.sizeof.bool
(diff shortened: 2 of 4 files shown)


doc/cffi-manual.texinfo
```diff
@@ -1940,6 +1940,11 @@ default.
 (convert-from-foreign 1 :boolean) @result{} t
 @end lisp
 
+@ForeignType{:bool}
+
+The @code{:bool} type represents the C99 @code{_Bool} or C++
+@code{bool}. Its size is mostly 1 byte except on OSX where it's an @code{int}.
```
luismbo: s/mostly/usually or "most often" perhaps?

tests/misc-types.lisp
```diff
@@ -86,6 +86,12 @@
           (bool-xor nil nil))
   (t t t t nil t nil))
 
+(defcfun "sizeof_bool" :unsigned-int)
+
+(deftest misc-types.sizeof.bool
+    (sizeof-bool)
+  #.(foreign-type-size #+darwin :int
+                       #-darwin :char))
```
luismbo: This should read `(foreign-type-size :bool)`, I think.

sionescu:
Fixed.

luismbo:
A test for the actual conversion would be nice.

sionescu:
The conversion test is done for :boolean, since :bool just wraps :boolean. Fixed foreign-type-size call too.

luismbo:
Well, that's an implementation detail that the tests shouldn't assume. ;-)

sionescu:
Done

luismbo:
Pushed 1da02a614f248bdf2c8b7f2ff8a87a4d70b0280f with minor tweaks. Thanks! :)

sionescu:
Can we make a bugfix release now so that it will be available in the next Quicklisp distribution ? Xach should make a new release next week.

luismbo:
Yes, this weekend.

PR #66, opened 2015-07-30
Travis: test only master branch

sionescu:
This was a feature branch doesn't get tested twice in a row: once as branch, once as pull request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+branches:
+  only:
+    - master
+
 language: lisp
 
 env:


luismbo:
Hmm, why do you push these branches to cffi/cffi instead of sionescu/cffi. I think the latter would be more idiomatic in the GitHub workflow?

sionescu:
More idiomatic when there's only one person with commit access to the main repository. With multiple committers it makes more sense for them to create feature branches and to test only pull requests and the integration branches(master/develop).

luismbo:
Hmm, OK. :) Go ahead and push / cherry-pick this onto master. :+1:

sionescu:
Done

PR #67, opened 2015-07-30
Enable TravisCI build on OSX, but allow it to fail

sionescu:


luismbo:
Hmm, a bit premature since cl-travis doesn't support osx yet, right?

luismbo:
And, come to think of it, with your recent plans to replace cl-travis with built-in lisp support in travis-ci itself, I don't think cl-travis will ever support OS X.

PR #68, opened 2015-08-01
few bricks for ecl

dkochmanski:
uffi and foreign function calling fixes

### DIFF
--- a/grovel/invoke.lisp
+++ b/grovel/invoke.lisp
@@ -57,6 +57,13 @@
               (integer   ret))
             "<see above>")))
 
+(defun process-output (process-stream)
+  (with-open-stream (process-stream process-stream)
+    (with-output-to-string (str)
+      (loop for char = (read-char process-stream nil)
+            while char
+            do (write-char char str)))))
+
 #+ecl
 (defun %invoke (command arglist)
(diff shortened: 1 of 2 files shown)


sionescu:
Is UFFI part of the external API of ECL ? Because in that case it was never able to load UFFI the library which is still used by a few packags. If at all possible I would suggest renaming the built-in UFFI package in ECL.

dkochmanski:
Hey! Yes, it is - but tought about changing this part of api in next release has crossed my mind. ECL UFFI is compatible with UFFI (but not 2.0, I have to implement remaining functions or remove it from api). I'll modify pull request for now to exclude "uffi: don't create nickname for ecl", and when test how things work and decide, I'll eventually make PR with just this patch (or not).

btw on irc my nick is jackdaniel

dkochmanski:
If I'll decide to preserve current api, will you accept this single patch? Some libraries depend on cffi-uffi-compat, and this breaks builds (package name conflict). clsql for example tries to find-package uffi, and if it doesn't find it, makes dependency - so it doesn't break on ECL (well, it breaks later, because of lack of functions from UFFI 2.0).

dkochmanski:
ok, closing this then.

PR #69, opened 2015-08-04
cffi-uffi-compat: ECL compability change

dkochmanski:
minimal patch for ECL support

### DIFF
--- a/uffi-compat/uffi-compat.lisp
+++ b/uffi-compat/uffi-compat.lisp
@@ -525,12 +525,12 @@ library type if type is not specified."
   #+clisp (sys::getenv (string var))
   #+(or cmu scl) (cdr (assoc (string var) ext:*environment-list* :test #'equalp
                              :key #'string))
-  #+gcl (si:getenv (string var))
+  #+(or ecl gcl) (si:getenv (string var))
   #+lispworks (lw:environment-variable (string var))
   #+lucid (lcl:environment-variable (string var))
   #+(or mcl ccl) (ccl::getenv var)
   #+sbcl (sb-ext:posix-getenv var)
-  #-(or allegro clisp cmu scl gcl lispworks lucid mcl ccl sbcl)
+  #-(or allegro clisp cmu ecl scl gcl lispworks lucid mcl ccl sbcl)
   (error 'not-implemented :proc (list 'getenv var)))
 
 ;; Taken from UFFI's src/os.lisp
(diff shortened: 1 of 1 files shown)


luismbo:
Interesting. Doesn't the package nickname `:uffi` collide with ECL's ffi package?

dkochmanski:
Yes, but from next release ecl wont have nickname uffi for ffi package. It doesn't implement api for uffi v. >= 2.0 and also its not asdf system, yet many packages put dep on uffi what fetches cffi version signalling conflict, so I've decided to piggyback on cffi compat - native ffi for ecl will be accessible as ffi package.

uffi-compat/uffi-compat.lisp
```diff
@@ -628,6 +632,12 @@ output to *trace-output*.  Returns the shell's exit code."
           `((with-foreign-strings ,(cdr bindings) ,@body))
           body)))
 
+#+ecl
+(defun foreign-string-length (foreign-string)
+  (declare (inline ffi:foreign-string-length))
+  (ffi:foreign-string-length foreign-string))
```
luismbo: I remember this function used to return a form instead of an integer, does it still do that?

uffi-compat/uffi-compat.lisp
```diff
@@ -525,24 +525,22 @@ library type if type is not specified."
...
+(defun run-shell-command (control-string &rest args)
   "Interpolate ARGS into CONTROL-STRING as if by FORMAT, and
 synchronously execute the result using a Bourne-compatible shell, with
 output to *trace-output*.  Returns the shell's exit code."
-  (unless output
-    (setq output *trace-output*))
-
-  (let ((command (apply #'format nil control-string args)))
+  (let ((command (apply #'format nil control-string args))
+        (output *trace-output*))
```
luismbo: Hmm. It seems like this never worked properly, right? When :output was passed, it would include that in the command string! That's what you're fixing here, correct?

dkochmanski: Right, forgot to mention, sorry

uffi-compat/uffi-compat.lisp
dkochmanski: Yes

luismbo: It's so weird. Have you ever seen it in use? I was wondering if we should leave it unimplemented to maintain the same behavior across implementations. What do you think?

dkochmanski: Ecl seems to use these when its compiled w/o dffi support. It crossed my mind to not add it there - if nobody lack this, then its ok with me - its accessible via native ffi anyway

luismbo: OK, I'd remove it then. Can you rebase the PR with that change? (Maybe add a note about the run-shell-command fix, if you're feeling chatty. :-))

dkochmanski:
bump

luismbo:
Merged, thanks! :+1:

PR #71, opened 2015-08-11
clisp: fix %invoke output

dkochmanski:
note, that temporary filename isn't random. its not in "#P/tmp" to keep it compat with windows


dkochmanski:
randomized file a bit

luismbo:
Great, thanks! :)

PR #73, opened 2015-09-24
Make CFFI-grovel play well with ASDF 3.1.6

fare:
Create linkable object files and include them in output-files for
compile-op, so that they can be found and linked by a recent-enough
ASDF (3.1.5.12 or later), allowing for delivery of a system with a
single static and/or dynamic library, and in the future, delivery
of a standalone executable that statically links required extensions.


fare:
OK, so I hadn't updated all the cflags handling infrastructure. Trying again...

fare:
I tweaked my patch a little bit, and added support for statically linked executable delivery on SBCL. This requires a slight patch to SBCL: https://bugs.launchpad.net/sbcl/+bug/1500628

fare:
If you install both these patches, and download the latest ASDF (currently 3.1.5.17; 3.1.5 is not enough), then you can:

```
(require "asdf")
(asdf:make 'cffi-grovel)
(asdf:operate 'cffi-grovel::static-image-op :osicat)
```

and you can run the executable at `(asdf:output-file 'cffi-grovel::static-image-op :osicat)`
and it will include osicat wrapped C functions such as

```
(osicat-posix:get-errno)
```

fare:
Oh, and this patch also fixes a concurrency bug in cffi — two builds in parallel would step on each other's toes, but now that's atomic.

fare:
Static linking also working of CLISP, though it's gross and not very portable at this point. I'd ask help from the CLISP maintainers to clean it up, but CLISP is currently unmaintained. Still works better than not at all.

fare:
Ping?

I updated this series of patches for a simpler more portable SBCL linkkit with a sbcl.o rather than libsbcl.a.

fare:
Support for CMUCL, SBCL, CLISP. The feature is basically working and ready for primetime.

luismbo:
Nice work, Faré. Sorry for not being very responsive. The feature is very interesting and all I have to contribute is minor bikesheding (e.g., mixed feelings about the `nest` macro). I think we're just missing User Manual documentation and at least a simple smoke test.

fare:
I rebased my branch.

Can you guide me: which sections should I modify and/or where should I add a new section in the manual?

luismbo:
https://common-lisp.net/project/cffi/manual/html_node/Groveller-ASDF-Integration.html#Groveller-ASDF-Integration is where the current functionality is documented. Perhaps the new features deserve sibling sections.

sionescu:
Hi @fare, nice work. I would prefer to see this condensed in a couple of commits, without the intermediate cleanups.

fare:
OK. I got the documentation and tests. What's left before you can merge?

Stelian: I can trivially merge everything into one commit. If you want "a couple" of commits, please tell me along what criteria to split the changes.

fare:
Also — I may have to add backward compatibility with old versions of ASDF. How far back should I go in providing compatibility?

fare:
OK, here it is as a single commit, with support hopefully back to ASDF 3.1.2, maybe further back in 3.0.x but I haven't tested.

sionescu:
@luismbo, how about we drop support for ASDF < 3.1, since it has been bundled with SBCL for some time already ?
@fare can you look into why this fails the tests on TravisCI ?

fare:
Failures on Travis are due to insufficient backward-compatibility support with versions older than 3.1.6.

Yes, dropping support for anything older than 3.1.2 would be awesome!

3.1.2 is more than a year old, and only CLISP (currently unmaintained) still ships with 3.0.x. ASDF comes with a script to overwrite CLISP's ASDF with a more recent one.

fare:
Actually, in replacing the Makefile with ASDF code, I was doing the wrong thing wrt shared object files used while testing. Trying again...

fare:
OK, passes tests again, now.

fare:
Maybe I should move the asdf-compat, c-toolchain and static-link files to their own system, asdf-c or so.

Pro: this allows this functionality to be used even without CFFI-grovel; if done with an appropriate API, this shields CFFI from evolution of this part of the code;

Con: this means CFFI depends on one more package; two more if I start using lisp-invocation, as I'm wont to do.

luismbo:
Let's try and merge this into CFFI and see how it goes. Do you feel confident about the current state of this pull request?

fare:
The CMUCL maintainer tells me that #+cmucl has been the preferred feature since 2011.

Yes, that's relatively recent in CMUCL history; yet it's old enough that it can be relied on.

I'm confident that the current code works on the platforms I tested it with, but
1- maybe some of it should be moved to some asdf-c extension that cffi would depend on.
  If so, it's better to not merge yet.
2- the SBCL extension required for the static-link feature has been applied then rolled back,
  because it didn't work on Windows, MacOS X, and Linux 32-on-64. In the latter case, I believe
  I needed a LDFLAGS = -m elf32-i386 which suggests CFFI would need the same (sigh).
3- none of these issues is a regression.

luismbo:
Point (1) makes sense. Perhaps including that extension in the cffi repo would be a good compromise? (But I don't object incorporating this into cffi-grovel as is.)

Regarding (2), is there any advantage in waiting for that issue to be settled?

fare:
Yes, leaving the extension in a CFFI repo is a good compromise for now. I'm still hesitating about its name: asdf-c ? asdf-linking ? asdf-toolchain ? cffi-toolchain ? c-toolchain ?

luismbo:
What'll be the scope of the new system?

fare:
The scope would be in my current patch, these files (currently under grovel/):
asdf-compat.lisp — supporting ASDF back to 3.1.2
c-toolchain.lisp — compiling and linking C code
static-link.lisp — building statically linked images and applications

fare:
OK, this new version creates a system and a package cffi-toolchain.

fare:
And yes, slightly more work will be required for portability to more platforms, but it should already be a much cleaner solution, with additional features on CLISP and CMUCL (and soon to be SBCL if I debug it).

No regression, slightly enhanced tests, but note that it depends on ASDF 3.1 now (which is over a year old).

cffi-toolchain.asd
```diff
@@ -0,0 +1,45 @@
+;;;; -*- Mode: lisp; indent-tabs-mode: nil -*-
+;;;
+;;; cffi-grovel.asd --- ASDF system definition for cffi-grovel.
+;;;
+;;; Copyright (C) 2007, Luis Oliveira  <loliveira@common-lisp.net>
```
luismbo: This should say cffi-toolchain rather than cffi-grovel. And I didn't write it, so I guess your name should be there? (We should remove these useless headers at some point in favor of a single LICENSE file, perhaps, but that's for another time.)

doc/cffi-manual.texinfo
```diff
@@ -6633,14 +6637,55 @@ clash with, say, CL symbols.}
...
 The internal package is created by Lisp code output from the C program
 written by @cffi{}-Grovel; if your specification file is
-exampleint.lisp, the exampleint.cffi.lisp file will contain the
+@file{exampleint.lisp}, the @file{exampleint.cffi.lisp} file will contain the
 @cffi{} definitions needed by the rest of your project.
 @xref{Groveller Syntax}.
 
-@node Groveller Implementation Notes,  , Groveller ASDF Integration, The Groveller
+@node Groveller Static Linking, Groveller Implementation Notes, Groveller ASDF Integration, The Groveller
+@section Static Linking
```
luismbo: Perhaps this can go onto a section outside the Groveller chapter?

fare:
OK, here's a new version.

fare:
Ping?

I don't think it's worth adding a Copyright Google for those small pieces I contributed; I believe there should be just one copyright file, and otherwise all files may just mention the authors refering to an AUTHORS file, and the license via its SPDX code: http://spdx.org/licenses/

luismbo:
Merged, thanks!

PR #74, opened 2015-10-11
cffi-grovel: Don't pass -m64 to GCC on ARM64.

stassats:
It doesn't accept such an argument.


luismbo:
Pushed b1eabb9 with minor whitespace tweak and a quote for consistency, I guess. Thanks.

PR #75, opened 2015-10-13
Bitfield groveller accepts defctyped-type

guicho271828:
https://bugs.launchpad.net/cffi/+bug/1505321


sionescu:
Thanks, I merged with a small modification

guicho271828:
I now noticed an invention of a wheel. canonicalize-foreign-type should do the job. Is that the modification?

sionescu:
Yes :)

guicho271828:
:+1:

PR #76, opened 2015-11-20
Add a bitfield enum grovel syntax.

tmccombs:
like bitfield but gets values form c enum rather than macro definitions.

### DIFF
--- a/doc/cffi-manual.texinfo
+++ b/doc/cffi-manual.texinfo
@@ -6585,6 +6585,14 @@ symbol as name, or a list @code{(name &key base-type)}.  For example:
     :optional t))
 @end lisp
 
+@deffn {Grovel Form} bitfield-enum name-and-opts &rest elements
+
+Defines a bitfield from a c enum as @code{((lisp-name &optional
+c-name) &key documentation)}. It works very similarly to
+@code{bitfield} but retrieves values from a c enum rather than macro
+definitions.
+@end deffn
+
 
 @c ===================================================================
 @c SECTION: Groveller ASDF Integration
--- a/grovel/grovel.lisp
+++ b/grovel/grovel.lisp
@@ -690,6 +690,33 @@ string."
           (format out "~&#endif~%"))
         (c-format out ")")))
     (c-format out ")~%")))
+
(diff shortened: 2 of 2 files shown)


luismbo:
This code is identical to `cenum`, right? Can we refactor things a little bit so that they share code?

Also, I wonder if we can merge `constant-bitfield` with `bitfield` (and perhaps `cenum` with `constantenum`, deprecating the latter) somehow? Ideally one wouldn't have to specify whether enums or pre-processor constant are being grovelled (maybe @sionescu knows how to do it!). If not, perhaps we can specify that (`enum` or `cpp-constant`) as option to `bitfield` (and `cenum`). What do you think? Too much trouble?

(Note to self, this has been discussed earlier this year <http://thread.gmane.org/gmane.lisp.cffi.devel/2805> and that thread should probably updated with the conclusions we reach in this pull request.)

tmccombs:
I can think of a couple of ways to avoid having to specify whether to grovel enums or pre-processor constant:

1. get rid of the #if conditions and just use the values. This would mean that you could no longer use optional or alternative names for pre-processor macros.
2. Unless `optional` is true, put the first cname in the #else clause. This would mean that unless OPTIONAL was true, if the constant wasn't defined you get an error rather than a warning, and if OPTIONAL was true, it would _have_ to be a pre-processor constant, not an enum value.

I'm not sure which would be better, or if either are better than having an option to bitfield or cenum.

tmccombs:
Also, cenum and bitfield-enum will work if the values are defined as pre-processor constants. whereas constantenum and bitfield don't work for enums. I think it might be better to change the names so that bitfield works for either and change the current bitfield to constantbitfield, but then it would break backwards compatibility for projects that depend on  bitfield supporting OPTIONAL and alternate CNAMEs.

guicho271828:
would you also improve the documentation of other similar constructs? I often feel I still don't understand the difference of e.g. constant, cenum, constantenum, and bitfield. [The doc](https://common-lisp.net/project/cffi/manual/html_node/Groveller-Syntax.html#Groveller-Syntax) describes how to use them, but it doesn't describe the difference and when to use them.

guicho271828:
ping

PR #77, opened 2015-11-21
Foreign octets

tmccombs:


tmccombs:
It looks like with-pointer-to-vector-data actually does half of what I want here. The other half would be a way to convert a pointer and size to a vector.

luismbo:
@tmccombs Have a look at `lisp-array-to-foreign`, `foreign-array-to-list` and the other surrounding functions. Do they do what you need?

tmccombs:
Looks like they do. Unfortunately they aren't documented in the manual (at least I couldn't find any documentation).

tmccombs:
@luismbo those functions are exactly what I want. But they are undocumented, unexported, and don't seem to work, because of a call to FOLLOW-TYPEDEFS, which reduces the FOREIGN-ARRAY type to a FOREIGN-POINTER type, and then the call to ELEMENT-TYPE fails. My guess is the intention was for follow-typedefs to reduce typedefs to a FOREIGN-POINTER. I'll try to see if I can get them working.

luismbo:
I felt a strange a tingling when looking at <https://github.com/cffi/cffi/commit/f8f7ac883a708febf1032574f519e36e27bbe6d2>. Maybe that's the commit that broke it?

luismbo:
@tmccombs the good news is that perhaps your efforts can be salvaged by adapting your code and documentation to foreign-array? :-)

tmccombs:
> I felt a strange a tingling when looking at f8f7ac8. Maybe that's the commit that broke it?
Probably.

luismbo:
Superseded by pr #79.

PR #78, opened 2015-11-21
Fix invoke to return output.

tmccombs:
Commit 2de8ece51b20ec0da177f8e38febb3127b1dafc2 changed invoke so that
it didn't return the output of running the command. This broke
pkg-config-cflags because that needs the output of running the
pkg-config command. This fixes that problem.

### DIFF
--- a/grovel/grovel.lisp
+++ b/grovel/grovel.lisp
@@ -277,7 +277,9 @@ int main(int argc, char**argv) {
                     (format *debug-io* "~&Attempting to continue anyway.~%")
                     (return)))))
       (appendf *cc-flags*
-               (parse-command-flags (invoke "pkg-config" pkg "--cflags"))))))
+               (parse-command-flags (run-program (list "pkg-config" pkg "--cflags")
+                                                 :output :string
+                                                 :error-output :interactive))))))
 
 ;;; This form also has some "read time" effects. See GENERATE-C-FILE.
 (define-grovel-syntax in-package (name)


luismbo:
Cool, thanks. I think it'd be useful if `invoke` still printed the program's output to *debug-io*. What do you think?

tmccombs:
Sure, I'll add that

tmccombs:
It looks like one of the builds failed to download asdf or something.

luismbo:
@fare, why did `invoke` stop signalling an error when the invoked command fails? (Meh, I failed at properly reviewing your changes to cffi-grovel. :-()

fare:
This is the WRONG thing.

In most cases, with `invoke`, you want the output to be on the console, for status and/or debugging. The previous `invoke` was really bad.

If you depend on getting output, please use directly `uiop:run-program` with `:output :string`

fare:
uiop:run-program already throws its own uiop:subprocess-error when the process fails. I didn't feel like wrapping it in a grovel-error.

luismbo:
@fare OK, and with this solution (that @tmccombs just implemented, thanks! :+1:) where does stderr go?

Also, `invoke` had the advantage of placing the output (namely the output to stderr) in the error message which I think is pretty helpful. Can we make the new invoke do that?

@avodonosov (I hope I got the right Anton), you depend on `grovel-error` right?

fare:
Where do you want the error output to go? Use `:error-output` to tell `uiop:run-program`. The old `invoke` was braindead. Please do not resurrect braindead non-portable interfaces, when someone spent months developing a portable non-braindead interface.

There is no one-size fits all that `invoke` could possibly provide. That's why `uiop:run-program` has all these keyword arguments. Don't reinvent it badly.

fare:
If you want to capture stdout and stderr and error status, use `:ignore-error-status` to prevent `uiop:run-program` from signalling its own `uiop:subprocess-error` then signal your own error.

luismbo:
@fare no worries. No plans to revive the old `invoke`, just trying to restore the good parts of the previous behaviour.

fare:
Thanks.

`invoke` was good enough in the bad old days before uiop:run-program existed. But it just didn't cut the mustard anymore.

So yes, if you want to preserve error output and send it to the console, I suggest `:error-output :interactive`.

avodonosov:
> @avodonosov (I hope I got the right Anton), you depend on grovel-error right?

@luismbo probably not right the Anton - I don't remember depending on `grovel-error` and haven't discussed this recently.

luismbo:
@akovalenko I think this is it: https://github.com/cl-test-grid/cl-test-grid/blob/8aff9632d597cbcb51d48c5e6f2089045a64a4ae/reporting/list-results.lisp#L139

avodonosov:
@luismbo , ah, yes! Thanks. Now I remember, it was me who asked to introduce grovel-error, right??

It's convenient for cl-test-grid to be able to distinguish grovel-error, because it often happens (e.g. the test machine doesn't have the foreign dev- library installed). With grovel-error I know it's not a regression is lisp library, but most likely a missing foreign dependency.

luismbo:
@tmccombs I've pushed dd0a078, that incorporates your fix and restores `grovel-error`. Please let me know if I fumbled anything. I hope to get to your other pull request soon!

PR #79, opened 2015-11-23
Get array operations working and exported.

tmccombs:
I haven't written documentation for this, at least not yet. But it seems to be working now.

### DIFF
--- a/src/early-types.lisp
+++ b/src/early-types.lisp
@@ -312,7 +312,7 @@ Signals an error if FOREIGN-TYPE is undefined."))
   ())
 
 (defun follow-typedefs (type)
-  (if (typep type 'foreign-type-alias)
+  (if (typep type 'foreign-typedef)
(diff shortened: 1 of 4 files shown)


src/early-types.lisp
```diff
@@ -312,7 +312,7 @@ Signals an error if FOREIGN-TYPE is undefined."))
   ())
 
 (defun follow-typedefs (type)
-  (if (typep type 'foreign-type-alias)
+  (if (typep type 'foreign-typedef)
```
luismbo: This breaks whatever f8f7ac883a708febf1032574f519e36e27bbe6d2 was trying to fix, so I've added a note to https://bugs.launchpad.net/cffi/+bug/1474631 about it.

luismbo:
Pushed 10198d3 with minor tweaks. Thanks!

PR #83, opened 2015-12-22
Add a cffi:*errno*

attila-lendvai:
What do you think about something like this?

### DIFF
--- a/src/errno.lisp
+++ b/src/errno.lisp
@@ -1,0 +1,64 @@
+;;;; -*- Mode: lisp; indent-tabs-mode: nil -*-
+;;;
+;;; errno.lisp --- Access to the OS errno.
+;;;
+;;; Copyright (C) 2015, Ryan Pavlik <rpavlik@gmail.com>
+;;; Copyright (C) 2015, Attila Lendvai <attila@lendvai.name>
+;;;
+;;; Permission is hereby granted, free of charge, to any person
+;;; obtaining a copy of this software and associated documentation
+;;; files (the "Software"), to deal in the Software without
+;;; restriction, including without limitation the rights to use, copy,
+;;; modify, merge, publish, distribute, sublicense, and/or sell copies
+;;; of the Software, and to permit persons to whom the Software is
+;;; furnished to do so, subject to the following conditions:
+;;;
+;;; The above copyright notice and this permission notice shall be
+;;; included in all copies or substantial portions of the Software.
+;;;
+;;; THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+;;; EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+;;; MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+;;; NONINFRINGEMENT.  IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+;;; HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+;;; WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+;;; OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+;;; DEALINGS IN THE SOFTWARE.
+;;;
+
+(in-package #:cffi)
+
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  #+(or linux freebsd)
+  (progn
+    (defcfun (%errno-location "__errno_location") :pointer)
+    (defcfun (%strerror "strerror") :string (errno :int)))
+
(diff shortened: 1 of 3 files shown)


attila-lendvai:
rebased, plus a little change in the wording

guicho271828:
seems good, it is implemented justly as a symbol macro.

luismbo:
There are some issues with Allegro CL. See <https://github.com/cffi/cffi/pull/31> and the mailing list discussion linked from there.

attila-lendvai:
oh my, what a mess!

i've read parts of that mailing list thread, and i have to admit i wasn't aware of many of those things... so take this patch with a piece of salt then.

attila-lendvai:
and on a second thought, maybe they should be called ```*posix-errno*``` and ```posix-strerror```?

luismbo:
It'd be interesting to understand what the failure modes of your approach are. It seems like your approach just _just work_ and I don't quite understand/remember why Allegro CL needs a more complicated mechanism (maybe the other Lisps have the same difficulties but ignore them?).

attila-lendvai:
well, first and foremost, is it part of the FFI contract that CL implementations won't call any syscalls (or backup/restore errno)? i mean, it's out of the ordinary to do so, but what if allegro does some bookkeeping that involves overwriting errno?

other than that, any sane OS should make errno a thread-local, so if we ignore the above then i don't see any other obvious issues besides aesthetics (what should the names be? shall we include an errno enum, or just rely on groveling? (my vote is on the latter)).

luismbo:
Yes, Allegro probably does something strange like that. Do you it's a showstopper? I can contact Franz's support to know more details.

Going back to your previous question about a `posix-` prefix. I believe `errno` is standard C.

An errno enum (or whatever) seems like it should be done via grovelling, yeah. (It's what Osicat does, for instance.)

attila-lendvai:
well, i personally don't care about allegro, because i don't even have a licence. and people can always not load/use the cffi/errno system on allegro, and we could even #-allegro it, or #+allegro some error informing them until the issue is resolved... if it affects this code, too. could anyone check?

oh, right, errno is a C standard (https://en.wikipedia.org/wiki/Errno.h), so definitely no posix- prefix.

then, as for me, i don't know about any showstoppers for the inclusion of this change.

@sionescu may have something to say?

src/errno.lisp
```diff
@@ -0,0 +1,64 @@
...
+;;; DEALINGS IN THE SOFTWARE.
+;;;
+
+(in-package #:cffi)
+
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  #+(or linux freebsd)
+  (progn
+    (defcfun (%errno-location "__errno_location") :pointer)
+    (defcfun (%strerror "strerror") :string (errno :int)))
```
luismbo: Shouldn't we be using the thread-safe version `strerror_r`? If so, then beware of differences between GNU and non-GNU versions of strerror_r. See https://github.com/osicat/osicat/commit/ea4c363135cf7c2a5e5ab69f7947ecb4168a2438.

src/package.lisp
```diff
@@ -175,4 +175,7 @@
    #:defcvar
    #:get-var-pointer
    #:foreign-symbol-pointer
+   ;; *errno* and strerror are only available when the :cffi/errno ASDF system is loaded
```
luismbo: What are the cons of including *errno*/strerror directly without an auxiliar ASDF system?

luismbo:
BTW, did you see the implementation of errno handling in Osicat? It's pretty neat and it includes a return type that automatically checks errno and signals a condition with the appropriate error message. :-)

src/errno.lisp
attila-lendvai: because i didn't even consider that someone could implement strerror, a function that returns constant strings, in a non-threadsafe way. i'll look around in osicat and update the PR.

src/package.lisp
attila-lendvai: when loading errno.lisp on a system that is not supported, it signals a continuable error at compile time and a hard error at runtime. we could include it in the main CFFI system and change the compile time signal to a warning.

sionescu:
I suggest only implementing the return of errno as second value in foreign-funcall and defcfun. Even that isn't easy because of OSX, where AFAIK there's no errno symbol that one can lookup. As for Windows, there's no single errno concept there, and they're not even a memory location but functions to call to retrieve the last error, e.g. WSAGetLastError() for the WinSock API.

sionescu:
I'm so tempted to port the cmucl syscall library to SBCL and make IOlib work only on SBCL/Linux/x86_64.

attila-lendvai:
ok, my current stance: i won't pursuit the inclusion of errno, because it's a bigger can of worms than i can handle. as for me, feel free to close this PR.

(i'll just move a basic errno code into the project i'm working on)

luismbo:
Valiant effort @attila-lendvai. :-)

attila-lendvai:
well, you know what i think about backwards compatibility @luismbo ... :)

whoever has energy to cater for silly ideas from the eighties, be my guest. meanwhile i'll kludge it and hide it behind an API that works for the relevant cases, and go on creating something better (or at the very least try to).

PR #84, opened 2016-01-07
Add automatic generation of CFFI bindings from c2ffi generated spec files

attila-lendvai:
controversial issues:

 - it incompatibly changes defcenum to be closer in semantics to C enums. IIRC not all of this is strictly necessary, but missing some of this would cause the generation to become more complicated (although for now i didn't ponder too much about what it would look like based on the old CFFI enum semantics).

 - the source went into the src/c2ffi directory, as opposed to all the other extensions. it also uses the cffi/c2ffi asdf feature and includes the system in the main cffi.asd file. i propose to later on change the other extensions to also follow this pattern.

example projects: https://github.com/attila-lendvai/hu.dwim.sdl (this one uses cl-ppcre regexps), and a simple one with only one binding and no cl-ppcre: https://github.com/attila-lendvai/hu.dwim.bluez

let me know if you see anything unaesthetic or disagreeable in there! i'd like to get this into the next quicklisp release.


luismbo:
OK, next quicklisp release sounds like a good goal!

attila-lendvai:
fix the test on ccl.

attila-lendvai:
rebased, and a comment changed.

cffi.asd
```diff
@@ -65,11 +65,28 @@
      (:file "foreign-vars")
      (:file "features")))))
 
-(defmethod operation-done-p ((o test-op) (c (eql (find-system :cffi))))
-  nil)
-
-(defmethod perform ((o test-op) (c (eql (find-system :cffi))))
-  (operate 'asdf:load-op :cffi-tests)
-  (operate 'asdf:test-op :cffi-tests))
```
luismbo: This makes `(asdf:test-system :cffi)` not do anything. It's not such a great loss, but what was the goal here?

cffi.asd
```diff
@@ -65,11 +65,28 @@
...
+(defsystem :cffi/c2ffi
```
luismbo: You mention that this change _uses the cffi/c2ffi asdf feature and includes the system in the main cffi.asd file_. What feature is that?

libffi/funcall.lisp
```diff
@@ -121,4 +120,6 @@
 
 (setf *foreign-structures-by-value* 'foreign-funcall-form/fsbv-with-libffi)
 
+(pushnew 'cffi-features:struct-by-value *features*)
```
luismbo: Hmm. cffi-features is obsolete and has been so for more than 10 years now. I think we can remove it! ;-) Maybe `:cffi-libffi` would be a good name? Why exactly do you need this feature anyway? Doesn't ASDF provide some nice way of detecting the presence or absence of some library?

src/c2ffi/asdf.lisp
```diff
@@ -0,0 +1,162 @@
...
+;;; OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+;;; DEALINGS IN THE SOFTWARE.
+;;;
+
+(in-package #:cffi/c2ffi)
+
+(defclass c2ffi-file (cl-source-file)
+  ((package :initarg :package
+            :initform nil
+            :accessor c2ffi-file/package)
```
luismbo: [bikeshedding] is the pattern in these accessor names `class-name/slot-name`? I don't dislike it, but it's inconsistent with the `class-name-slot-name` (or just `slot-name` for internal stuff) convention used elsewhere in CFFI.

src/c2ffi/asdf.lisp
```diff
@@ -0,0 +1,162 @@
...
+generation for the other platforms is allowed to fail
+\(see ENSURE-SPEC-FILE-EXISTS).
+
+It's advisable to deliver these json files with the project, so that users
+don't need to have c2ffi installed.
+
+Once/if the json file is available for the underlying platform, then the json
+file is used to generate a lisp file with CFFI definitions (see
+PROCESS-C2FFI-SPEC-FILE). This file will then be compiled as any other lisp
+file, except that it's will be stored in the fasl cache."))
```
luismbo: This documentation will be a great start for a c2ffi chapter in the CFFI manual! :-)

src/c2ffi/asdf.lisp
```diff
@@ -0,0 +1,162 @@
...
+  (list (first (output-files 'generate-op c))))
+
+(defmethod component-depends-on ((op generate-op) (c c2ffi-file))
+  `((prepare-op ,c) ,@(call-next-method)))
+
+(defmethod component-depends-on ((op compile-op) (c c2ffi-file))
+  `((generate-op ,c) ,@(call-next-method)))
+
+(defmethod component-depends-on ((op load-source-op) (c c2ffi-file))
+  `((generate-op ,c) ,@(call-next-method)))
```
luismbo: These `component-depends-on` methods are saying that compilation/loading depend on generation which in turns depends on preparation, right?

src/c2ffi/asdf.lisp
```diff
@@ -0,0 +1,162 @@
...
+
+(defmethod output-files ((op generate-op) (c c2ffi-file))
+  (let* ((input-file (first (input-files op c)))
+         (spec-file (spec-path input-file))
+         (generated-lisp-file (make-pathname :type "lisp"
+                                             :defaults spec-file)))
+    ;; this order here influences various other places
```
luismbo: Does _this order here_ refer to the `list*`? Can you clarify this comment a little bit?

src/c2ffi/asdf.lisp
```diff
@@ -0,0 +1,162 @@
...
+                   ;; then that makes ASDF to unconditionally redo the rest of the
+                   ;; generation and compilation.
+                   (unless (probe-file spec-file)
+                     (list spec-file)))
+            t)))
+
+(defmethod perform ((op generate-op) (c c2ffi-file))
+  (let* ((output-file (first (output-files op c)))
+         (input-file (first (input-files op c)))
+         (spec-file (let ((*c2ffi-executable* (if (slot-boundp c 'c2ffi-executable)
```
luismbo: This binding pattern for `*c2ffi-executable*` and `*trace-c2ffi` is here so that they can be dynamically bound around the loading of a cffi/c2ffi-using library, is it?

src/c2ffi/c2ffi.lisp
```diff
@@ -0,0 +1,163 @@
...
+    (format *debug-io* "~&; Invoking: ~A~{ ~A~}~%" program args))
+  (zerop (nth-value 2 (uiop:run-program (list* program args) :output output :error-output error-output :ignore-error-status ignore-error-status))))
+
+(defun generate-spec-with-c2ffi (input-header-file output-spec-path &key arch sys-include-paths ignore-error-status)
+  "Run c2ffi on `INPUT-HEADER-FILE`, outputting to `OUTPUT-FILE` and
+`MACRO-OUTPUT-FILE`, optionally specifying a target triple `ARCH`."
+  (uiop:with-temporary-file (:pathname tmp-macro-file
+                             :keep *trace-c2ffi*)
+    nil
+    :close-stream
```
luismbo: First time I see this "label" API. It's interesting. :-)

src/c2ffi/c2ffi.lisp
```diff
@@ -0,0 +1,163 @@
...
+    (let* ((arch (when arch (list "-A" arch)))
+           (sys-include-paths (loop
+                                :for dir :in sys-include-paths
+                                :append (list "-i" dir))))
+      ;; Invoke c2ffi to emit macros into TMP-MACRO-FILE
```
luismbo: I'm missing the global picture here. What macros are these? Where do they come from?

src/c2ffi/c2ffi.lisp
```diff
@@ -0,0 +1,163 @@
...
+  (make-pathname :defaults base-name
+                 :name (strcat (pathname-name base-name)
+                               (if version
+                                   (strcat "-" version)
+                                   "")
+                               "."
+                               arch)
+                 :type "spec"))
+
+(defun find-local-spec (base-name &key (otherwise nil otherwise?))
```
luismbo: This `otherwise?` logic isn't being used, is it? We can remove it then, right?

src/c2ffi/c2ffi.lisp
```diff
@@ -0,0 +1,163 @@
...
+        (values h-name m-name)
+        (let ((local-arch (local-arch)))
+          (unless (c2ffi-executable-available?)
+            (error "No spec found for ~S on arch '~A' and c2ffi not found"
+                   header-file-path local-arch))
+          (generate-spec-with-c2ffi header-file-path
+                                    (spec-path header-file-path :arch local-arch :version version)
+                                    :arch local-arch
+                                    :sys-include-paths sys-include-paths)
+          ;; also run c2ffi for other architectures, but tolerate failure
```
luismbo: What's the goal of running c2ffi for multiple architectures? Is it help maintainers populate their library of pre-built specs?

src/c2ffi/c2ffi.lisp
```diff
@@ -0,0 +1,163 @@
...
+            (unless (or (string= local-arch arch)
+                        (member arch exclude-archs :test #'string=))
+              (unless (generate-spec-with-c2ffi header-file-path
+                                                (spec-path header-file-path :arch arch :version version)
+                                                :arch arch
+                                                :sys-include-paths sys-include-paths
+                                                :ignore-error-status t)
+                (warn "Failed to generate spec for other arch: ~S" arch))))
+          (or (find-local-spec header-file-path)
+              (error "Error finding spec for ~S after running c2ffi" header-file-path))))))
```
luismbo: This error never kicks in because of `find-local-spec`'s _otherwise_ logic, right?

src/c2ffi/generator.lisp
```diff
@@ -0,0 +1,755 @@
+;;;; -*- Mode: lisp; indent-tabs-mode: nil -*-
```
luismbo: NOTE TO SELF: continue review from here, tomorrow.

cffi.asd
attila-lendvai: ASDF treats '/' specially, the system 'foo/bar' can be put into the 'foo.asd' file and ASDF will still find it.

it's a way to group many related .asd files into one. i propose to convert the rest of the .asd's, too. backwardness can be achieved by retaining a simple cffi-libffi.asd with an empty system that depends on cffi/libffi.

libffi/funcall.lisp
attila-lendvai: i'm just keeping the old code alive here, and apparently i was wrong assuming that cffi-features is the way to go... :)

:cffi-libffi is not a good symbol, though, because this is about the ability to pass structures by value, and using libffi is just one way to implement that.

maybe :cffi/struct-by-value then?

or :cffi-fsbv? (i never liked fsbv because for me it's a pain to type and remember, and doesn't communicate anything to the person first approaching the codebase)

but i'm not using this at all... i rely on explicit dependencies, so as for me i'm fine with the complete removal of this from *features*. it will just lead to headaches anyways if someone compiles their code with cffi-libffi loaded, and then tries to reload the fasls later on into an image without cffi-libffi... the only sane way to deal with this (that i know of) is explicit ASDF dependencies.

src/c2ffi/asdf.lisp
attila-lendvai: yep. i prefer this, or the slot-name-of pattern to communicate that they are just accessors, but i don't have an emotional attachment to it. just let me know and i'll reshape it to match the rest of the code.

and to be honest a large part of that thing should be squeezed into a macrolet anyways... that would use slot-value and the accessors could be done away with altogether.

when deciding, keep in mind that this all is in a separate package.

src/c2ffi/asdf.lisp
attila-lendvai: sorry, i have no clue... :)

i've copied this from cffi-grovel which was written by @fare, AFAIK.

src/c2ffi/asdf.lisp
attila-lendvai: yes. or straight out setf the global binding before calling asdf:load-system.

src/c2ffi/c2ffi.lisp
attila-lendvai: IIRC that ```nil``` there is a workaround of a bug in an old UIOP. will note that in a comment...

src/c2ffi/c2ffi.lisp
attila-lendvai: wrote a longer comment, i'll rebase and push -f.

      ;; Invoke c2ffi to first emit C #define's into TMP-MACRO-FILE. First ask c2ffi
      ;; to generate a file of C global variables that are assigned the value of
      ;; the #define's, so that in the second pass below the C compiler
      ;; evaluates for us their bodies and thus we can get hold of their value.
      ;; This is a kludge and eventually we could/should support generating
      ;; cffi-grovel files, and in grovel mode not rely on this kludge.


        ;; Write a tmp header file that #include's the original input file and
        ;; the above generated macros file which will form the input for our
        ;; final, second pass.

src/c2ffi/c2ffi.lisp
attila-lendvai: it's used right here a few lines below.

i can just convert it to an ```&key error?``` or ```errorp``` or ```error-p```...

what's your preference?

FTR, this otherwise construct is crippled anyways... a full implementation is e.g. in this file: http://hub.darcs.net/hu.dwim/hu.dwim.util/browse/source/miscellaneous.lisp#249

src/c2ffi/c2ffi.lisp
attila-lendvai: the idea behind c2ffi is that users of the library should not need a functioning c2ffi executable and cffi/c2ffi just takes the appropriate .spec file that have been checked into the repo.

now, if for whatever reason we end up needing to (re)generate the .spec file for an arch, which means that we need a functioning c2ffi exe, then let's just try to generate the spec files for the rest of the architectures, too.

ultimately, not even this protects us against C libs that break ABI level compatibility. to deal with those (and e.g. inline function declarations in header files) the generator should support generating cffi-grovel files and go through the C compiler on each installation. maybe in a future version...

src/c2ffi/c2ffi.lisp
attila-lendvai: good catch, thanks!

src/c2ffi/asdf.lisp
attila-lendvai: i have zero experience with texinfo.

and i would also wait a little until the code settles down. for example it's unclear for me how important it is, from a practical perspective, to move on to support generating cffi-grovel files (see below for discussion).

src/c2ffi/generator.lisp
```diff
@@ -0,0 +1,755 @@
...
+
+(defun call-transformer-hook (hook &rest args)
+  (apply hook
+         ;; indiscriminately add one keyword arg entry to warn
+         (append args '(just-a-warning "Make sure your transformer hook has &key &allow-other-keys for future extendability."))))
+
+(defun find-cffi-type-or-die (type-name &optional (namespace :default))
+  (when (eq namespace :enum)
+    ;; TODO FIXME this should be cleaned up in CFFI. more about namespace confusion at:
+    ;; https://bugs.launchpad.net/cffi/+bug/1527947
```
luismbo: Agreed.

src/c2ffi/generator.lisp
```diff
@@ -0,0 +1,755 @@
...
+                            :external-format (asdf/driver:encoding-external-format output-encoding))
+        (let* ((*package* (or (find-package package-name)
+                              (make-package package-name)))
+               ;; Make sure we use an uninterned symbol, so that it's neutral to READTABLE-CASE.
+               (package-name (make-symbol (package-name *package*)))
+               ;; Let's rebind a copy, so that when we are done with the generation
+               ;; then the CFFI type repository is also reverted back to the
+               ;; previous state. This avoids future redefinition warning if they
+               ;; get added to defctype and friends.
+               (cffi::*type-parsers* (copy-hash-table cffi::*type-parsers*))
```
luismbo: At some point, we should export an API for this, but this is fine for now.

src/c2ffi/generator.lisp
```diff
@@ -0,0 +1,755 @@
...
+                                 (cffi:defctype ,(function-pointer-type-name) :pointer)))
+          (when (and foreign-library-name
+                     foreign-library-spec)
+            (when (stringp foreign-library-name)
+              (setf foreign-library-name (safe-read-from-string foreign-library-name)))
+            (output/code `(cffi:define-foreign-library ,foreign-library-name
+                            ,@foreign-library-spec))
+            ;; NOTE: maybe we should register this as an UIOP:REGISTER-IMAGE-RESTORE-HOOK?
+            ;; On some platforms DLL's need to be relinked after an image save and restart.
+            (output/code `(cffi:use-foreign-library ,foreign-library-name)))
```
luismbo: `define-foreign-library` registers the library, and it's possible to list all defined libraries, reopen them all, things like that. Right now, users will have to deal with that themselves, but the tools are there. If we're going to make it more automatic, then CFFI should be responsible for it, not CFFI/C2FFI, I suppose.

src/c2ffi/generator.lisp
```diff
@@ -0,0 +1,755 @@
...
+                                          +name-kinds+))
+               (*allow-pointer-type-simplification* allow-pointer-type-simplification)
+               (*allow-skipping-struct-fields* allow-skipping-struct-fields)
+               (*assume-struct-by-value-support* assume-struct-by-value-support)
+               (*ffi-name-transformer* (canonicalize-transformer-hook ffi-name-transformer))
+               (*ffi-type-transformer* (canonicalize-transformer-hook ffi-type-transformer))
+               (json (json:decode-json in)))
+          (output/string +generated-file-header+)
+          ;; some forms that are always emitted
+          (mapcar 'output/code `((defpackage ,package-name)
```
luismbo: Maybe use `mapc` to emphasize that the return value is not used? Also, that `defpackage` should have an empty `:use` list to ensure it's empty and not some other default list, right?

src/c2ffi/generator.lisp
```diff
@@ -0,0 +1,755 @@
...
+            (integerp value))
+       (make-define-constant-form name value))
+      ((and (member cffi-type '(:float :double))
+            (floatp value))
+       (make-define-constant-form name value))
+      ((member cffi-type '(:string (:pointer :char)) :test 'equal)
+       (make-define-constant-form name value))
+      (t
+       (warn "Don't know how to emit a constant of CFFI type ~S, with value ~S (json type is ~S)." cffi-type value type)
+       (values)))))
```
luismbo: I skimmed the processors a little bit... :)

src/c2ffi/integration/cl-ppcre.lisp
```diff
@@ -0,0 +1,36 @@
...
+
+(in-package #:cffi/c2ffi)
+
+(defun make-rule-matcher/cl-ppcre (pattern)
+  (let ((scanner (cl-ppcre:create-scanner pattern)))
+    (named-lambda cffi/c2ffi/cl-ppcre-rule-matcher
+        (string)
+      (funcall scanner string 0 (length string)))))
+
+(setf *make-matcher-from-rule-hook* 'make-rule-matcher/cl-ppcre)
```
luismbo: IIUC, the presence of this systems affects how the patterns are processed, right? Doesn't that mean that a system that was using plain strings might feel if for some reasons cffi/c2ffi+cl-ppcre is loaded?

Can't we just depend on cl-ppcre unconditionally and call it a day? (Hmm, does that mean we'd be requiring cl-ppcre just for loading a cffi/c2ffi-generated set of bindings? Can we avoid that somehow? More generally, can we avoid loading c2ffi altogether if the bindings are already generated? Hmm, this last bit probably doesn't make much sense. The idea is to cache [i.e., commit] specs, not the generated bindings, right?)

src/c2ffi/package.lisp
```diff
@@ -0,0 +1,40 @@
+;;;; -*- Mode: lisp; indent-tabs-mode: nil -*-
+;;;
+;;; package.lisp --- Groveler DEFPACKAGE.
```
luismbo: Groveler -> CFFI/C2FFI :-)

src/early-types.lisp
```diff
@@ -53,17 +53,67 @@
...
+
+(define-condition type-related-error (cffi-error)
+  ((type-name :initarg :type-name
+              :initform (error "Must specify TYPE-NAME.")
+              :accessor type-related-error/type-name)
+   (namespace :initarg :namespace
+              :initform :default
+              :accessor type-related-error/namespace)))
+
+(defun type-related-error/compund-name (e)
```
luismbo: compund -> compound

src/early-types.lisp
```diff
@@ -53,17 +53,67 @@
 (defvar *type-parsers* (make-hash-table :test 'equal)
   "Hash table of defined type parsers.")
 
+(define-condition cffi-error (error)
+  ())
+
+(define-condition type-related-error (cffi-error)
```
luismbo: This is was long overdue. `foreign-type-error` seems more inline with other CFFI names, maybe? Naming is hard...

src/early-types.lisp
```diff
@@ -53,17 +53,67 @@
...
+  (error 'undefined-type-error :type-name type-name :namespace namespace))
+
+;; TODO this is not according to the C namespace rules,
+;; see bug: https://bugs.launchpad.net/cffi/+bug/1527947
+(deftype c-namespace-name ()
+  '(member :default :struct :union))
+
+;; for C namespaces read: https://stackoverflow.com/questions/12579142/type-namespace-in-c
+;; (section 6.2.3 Name spaces of identifiers)
+;; NOTE: :struct is probably an unfortunate name for the tagged (?) namespace
```
luismbo: Mimicking the C namespace rules was not what this `namespace` thing was trying to do. Let's call it `:tagged`, no problem.

src/early-types.lisp
```diff
@@ -53,17 +53,67 @@
...
-      (if (eq namespace :default)
-          (error "unknown CFFI type: ~S." symbol)
-          (error "unknown CFFI type: (~S ~S)." namespace symbol))))
+      (undefined-type-error symbol namespace)))
 
 (defun (setf find-type-parser) (func symbol &optional (namespace :default))
   "Set the type parser for SYMBOL."
+  (check-type symbol (and symbol (not null)))
+  (check-type namespace c-namespace-name)
+  ;; TODO Shall we signal a redefinition warning here?
```
luismbo: How?

src/enum.lisp
```diff
@@ -27,6 +27,16 @@
 
 (in-package #:cffi)
 
+;; TODO the accessors names are rather inconsistent:
+;; FOREIGN-ENUM-VALUE           FOREIGN-BITFIELD-VALUE
+;; FOREIGN-ENUM-KEYWORD         FOREIGN-BITFIELD-SYMBOLS
+;; FOREIGN-ENUM-KEYWORD-LIST    FOREIGN-BITFIELD-SYMBOL-LIST
+;; I'd rename them to: FOREIGN-*-KEY(S) and FOREIGN-*-ALL-KEYS -- attila
```
luismbo: Maybe `-NAME(S)` instead of `-KEY(S)`?

src/enum.lisp
```diff
@@ -27,6 +27,16 @@
...
+
+;; TODO bitfield is a confusing name, because the C standard calls
+;; the "int foo : 3" type as a bitfield. Maybe rename to defbitmask?
+;; -- attila
```
luismbo: Aye.

src/enum.lisp
```diff
@@ -37,59 +47,138 @@
 ;;; constant for the type by a type translator when passed as
 ;;; arguments or a return value to a foreign function.
 
-(defclass foreign-enum (foreign-typedef enhanced-foreign-type)
+(defclass foreign-enum (named-foreign-type enhanced-foreign-type)
```
luismbo: Tell me more about this superclass change. :)

src/enum.lisp
```diff
@@ -37,59 +47,138 @@
...
+  '(and symbol (not null)))
+
+(defparameter +valid-enum-base-types+ '(:char :unsigned-char
+                                        :short :unsigned-short
+                                        :int :unsigned-int
+                                        :long :unsigned-long
+                                        :long-long :unsigned-long-long))
+
+(defun parse-foreign-enum-like (type-name base-type values
+                                &optional field-mode?)
```
luismbo: I know this is bikeshedding, but CFFI uses the, admittedly retro, `-p` convention. :)

src/enum.lisp
```diff
@@ -37,59 +47,138 @@
...
+  (let ((keyword-values (make-hash-table :test 'eq))
+        (value-keywords (make-hash-table))
+        (field-keywords (list))
+        (bit-index->keyword (make-array 0 :adjustable t
+                                        :element-type t))
+        (default-value (if field-mode? 1 0))
+        (largest-value 0))
     (dolist (pair values)
-      (destructuring-bind (keyword &optional (value default-value))
+      (destructuring-bind (keyword &optional (value default-value value-p))
```
luismbo: More bikeshedding, the `-p` convention rules go like this: `multi-word-p`, `wordp` (and `some-prefix-wordp`). ;)

src/enum.lisp
```diff
@@ -37,59 +47,138 @@
...
+                                  :initial-element nil)))
+            (setf (aref bit-index->keyword bit-index)
+                  keyword)))))
+    (if base-type
+        (progn
+          (setf base-type (canonicalize-foreign-type base-type))
+          (assert (member base-type +valid-enum-base-types+ :test 'eq) ()
+                  "Invalid base type ~S for enum type ~S. Must be one of ~S."
+                  base-type type-name +valid-enum-base-types+))
+        ;; details: https://stackoverflow.com/questions/1122096/what-is-the-underlying-type-of-a-c-enum
```
luismbo: Interesting. Did you actually come across an enum that doesn't fit an int?

src/enum.lisp
```diff
@@ -135,12 +224,12 @@
          (%foreign-enum-value ,type ,value :errorp t)
          ,value)))
 
-;;; There are two exansions necessary for an enum: first, the enum
+;;; There are two expansions necessary for an enum: first, the enum
 ;;; keyword needs to be translated to an int, and then the int needs
 ;;; to be made indirect.
 (defmethod expand-to-foreign-dyn-indirect (value var body (type foreign-enum))
```
luismbo: Off-topic: I'm still unconvinced about all these `-indirect` methods...

src/enum.lisp
```diff
@@ -152,99 +241,94 @@
...
 
 (defmacro defbitfield (name-and-options &body masks)
   "Define an foreign enumerated type."
-  (discard-docstring masks)
-  (destructuring-bind (name &optional (base-type :int))
-      (ensure-list name-and-options)
-    `(eval-when (:compile-toplevel :load-toplevel :execute)
-       (notice-foreign-type
-        ',name (make-foreign-bitfield ',name ',base-type ',masks)))))
+  (%defcenum-like name-and-options masks 'make-foreign-bitfield))
```
luismbo: Nicely done. :)

src/enum.lisp
```diff
@@ -152,99 +241,94 @@
...
+             (check-type symbol symbol)
+             (or (gethash symbol (keyword-values type))
+                 (error "~S is not a valid symbol for bitfield type ~S."
+                        symbol type))))
+    (declare (dynamic-extent #'process-one))
+    (if (consp symbols)
+        (reduce #'logior symbols :key #'process-one)
+        (process-one symbols))))
+
+(defmacro %foreign-bitfield-value/form (type symbols)
```
luismbo: Why is this a macro?

src/enum.lisp
```diff
@@ -152,99 +241,94 @@
...
       form))
 
 (defun %foreign-bitfield-symbols (type value)
   (check-type value integer)
-  (loop for mask being the hash-keys in (value-symbols type)
-            using (hash-value symbols)
-        when (= (logand value mask) mask)
-        append symbols))
+  (check-type type foreign-bitfield)
+  (loop
```
luismbo: [bikeshedding] LOOPs throughout CFFI are indented like this:

```lisp
(loop for x in foo
      do (bar)
      when (baz)
        collect (quux))
```

Different LOOP styles mixed together makes my OCD-sense tingle. :-/

src/enum.lisp
```diff
@@ -152,99 +241,94 @@
...
+    :with bit-index->keyword = (bit-index->keyword type)
+    :for bit-index :from 0 :below (array-dimension bit-index->keyword 0)
+    :for mask = 1 :then (ash mask 1)
+    :for key = (aref bit-index->keyword bit-index)
+    :when (and key
+               (= (logand value mask) mask))
+    :collect key))
+
+(defmacro %foreign-bitfield-symbols/form (type value)
```
luismbo: This probably shouldn't be a macro either. And it's capturing the type variable by chance in the error form, right?

src/types.lisp
```diff
@@ -30,6 +30,11 @@
 
 ;;;# Built-In Types
 
+;; NOTE: In the C standard there's a "signed-char":
+;; https://stackoverflow.com/questions/436513/char-signed-char-char-unsigned-char
+;; and "char" may be either signed or unsigned, i.e. treating it as a small int
+;; is not wise. At the level of CFFI we can safely ignore this and assume that
+;; :char is mapped to "signed-char" by the CL implementation under us.
```
luismbo: Indeed, I think we have some tests in the suite verifying that this is true.

src/types.lisp
```diff
@@ -659,6 +664,14 @@ The foreign array must be freed with foreign-array-free."
         offset
         (+ offset (- align rem)))))
 
+(defmacro with-tenative-type-definition ((name value namespace) &body body)
```
luismbo: tenative -> tentative

src/types.lisp
```diff
@@ -667,28 +680,30 @@ The foreign array must be freed with foreign-array-free."
...
-          (incf current-offset (* count (foreign-type-size type))))
-        (setq firstp nil))
-      ;; calculate padding and alignment
-      (setf (alignment struct) max-align) ; See point 1 above.
-      (let ((tail-padding (- max-align (rem current-offset max-align))))
-        (unless (= tail-padding max-align) ; See point 3 above.
-          (incf current-offset tail-padding)))
-      (setf (size struct) (or size current-offset))
-      (notice-foreign-type name struct :struct))))
+      (with-tenative-type-definition (name struct :struct)
```
luismbo: This is interesting and looks like a great idea. Can you tell me a bit more about what lead you to this solution?

tests/bindings.lisp
```diff
@@ -101,7 +101,8 @@
   (let ((*foreign-library-directories* (list (load-directory))))
     (load-foreign-library 'libtest)
     (load-foreign-library 'libtest2)
-    #+fsbv (load-foreign-library 'libfsbv)
+    (when (cffi-feature-p :struct-by-value)
+      (load-foreign-library 'libfsbv))
```
luismbo: I don't understand what that `#+fsbv` was doing there since `cffi-tests` depends on `cffi-libffi` unconditionally. I think we can just load libfsbv unconditionally.

tests/enum.lisp
```diff
@@ -64,12 +66,57 @@
...
+  (enum-size-one 1)
+  (enum-size-int #.(1- (expt 2 (1- (* (foreign-type-size :int) 8)))))
+  (enum-size-two 2))
+
+(defcenum enum-size.long
+  (:long #.(1- (expt 2 (1- (* (foreign-type-size :long) 8)))))
+  (:one 1)
+  (:two 2))
+
+#+nil ;; :long-long is often the same size as :long
```
luismbo: Why is this test commented out?

tests/fsbv.lisp
```diff
@@ -137,7 +137,7 @@
   5)
 
 ;;; returning struct with bitfield member (bug #1474631)
-(defbitfield (struct-bitfield :unsigned-int)
+(defbitfield (struct-bitfield :int)
```
luismbo: What's going on here?

tests/package.lisp
```diff
@@ -28,6 +28,6 @@
 (in-package #:cl-user)
 
 (defpackage #:cffi-tests
-  (:use #:cl #:cffi #:cffi-sys #:regression-test)
+  (:use #:cl #:cffi #:cffi-sys #:cffi-features #:regression-test)
```
luismbo: Like we've discussed before, let's drop cffi-features.

cffi.asd
luismbo: Sounds like a good idea. We can probably figure out a way to put some sort of deprecation warning when loading those empty systems.

libffi/funcall.lisp
luismbo: I suggest we just leave `:fsbv`, maybe with a deprecation note. (But perhaps @liamh feels differently.) We can definitely drop the `#+fsbv` from `cffi-tests` as I mentioned in another comment.

src/c2ffi/asdf.lisp
luismbo: Let's leave like this. If it bothers me too much in the future (unlikely), I can massage it myself instead of pestering you. ;)

src/c2ffi/asdf.lisp
luismbo: OK, I'll try and tackle the manual bits. I suppose it'll help me understand the code better. I've opened https://bugs.launchpad.net/cffi/+bug/1538781 to track that. I hope it doesn't sit there forgotten forever. :(

src/c2ffi/asdf.lisp
luismbo: Heh, OK.

luismbo:
Should we tackle the merging of enum and struct/union namespaces now or later?

You said that "ultimately here we need to decide whether to tailor cffi towards ffi's written by hand or generated ffi's". I didn't see anything particularly untailored towards FFIs written by hand. What did you mean? Did I miss something?

Anyway, great work! Thanks!

luismbo:
P.S.: did you notice the Travis test failures on 32-bit SBCL?

src/c2ffi/generator.lisp
attila-lendvai: unfortunately defpackage's behavior is undefined when the name already names an existing package. (will do the suggestions anyway, better than nothing)

src/c2ffi/integration/cl-ppcre.lisp
attila-lendvai: yep, but i was hoping that cl-ppcre is backwards compatible enough with a string search, but thinking again it's not. especially that the default rule matcher is an ends-with...

i'm not happy with the ppcre dependency, but i don't have a sane idea for an alternative either... will sleep on it and see if something comes up.

re checking in the generated bindings: this is actually not a bad idea, because the .spec files practically define the .lisp files (modulo cffi/c2ffi generator changes, or its parameters in the c2ffi-file in the asdf system).

trouble is that the user libs need to straight out :defsystem-depends-on cffi/c2ffi to be able to define the system itself, so a hard dependency on cl-ppcre would come with cffi/c2ffi.

maybe user libs could have two systems:
 - one with a full cffi/c2ffi dependency, and the ability to regenerate the binding
 - and another one that merely loads the binding, assuming that it has been checked into the repo.

i'll sleep one on this, too. i think it's actually a good idea to keep the generated lisp files also next to the .spec files, and let the users decide whether they check it into their repos or not.

src/c2ffi/package.lisp
attila-lendvai: damn, i got caught! :)

attila-lendvai: can i just delete that line altogether? it's a perfect example of harmful documentation...

src/early-types.lisp
attila-lendvai: yeah, i'm not happy with that either, but it's still better, so i'll rename them. this one is just a mixin, but this naming pattern is even worse in the later condition classes.

shall i get rid of the class-name/slot accessors?

src/early-types.lisp
attila-lendvai: i guess by checking if the hashtable value will be overwritten or assigned for the first time.

note: i may be unaware of some compile-time/load-time/execute headaches, though.

src/enum.lisp
attila-lendvai: the problem with -NAME is that enums themselves have names, so it should be -ENTRY-NAME(S) then, which is kinda long.

re plural: i'd use -KEY and -KEYS as needed.

src/enum.lisp
attila-lendvai: who remembers after such a long time...? :)

i think the problem was that ```follow-typedefs``` went all the way down to :int somewhere, which was wrong. an enum is not a typedef to an int, it's an enum...

src/enum.lisp
attila-lendvai: i thought it's a defun-local question mark and will slip through... :)

src/enum.lisp
attila-lendvai: nope.

i just bit the bullet and learned about the relevant parts of the C standard, because i was too confused at the beginning. and if i got to know this anyway, then i thought i'll just implement it in CFFI, too.

src/enum.lisp
attila-lendvai: good for you! i don't even know what they are... :)

i was missing a couple of words of documentation about these defgeneric based macro expand protocols. e.g. what is the intended difference between ```-translate``` and ```-convert```, etc.

now i think i have a partial picture, but it was very opaque when i first approached the codebase.

cffi.asd
attila-lendvai: it's a mistake, i have assumed that a functionality we use is in asdf, but it's in fact in our own lib... i'll revert this partially (the operation-done-p method is not needed anymore)

luismbo: Roger that.

libffi/funcall.lisp
attila-lendvai: ok, dropped that whole commit and wrote the following comment:

```
;; DEPRECATED Its presence encourages the use of #+fsbv which may lead to the
;; situation where a fasl was produced by an image that has fsbv feature
;; and then ends up being loaded into an image later that has no fsbv support
;; loaded. Use explicit ASDF dependencies instead and assume the presence
;; of the feature accordingly.
(pushnew :fsbv *features*)
```

src/c2ffi/asdf.lisp
attila-lendvai: thanks!

i've also subscribed to that bug.

src/c2ffi/asdf.lisp
attila-lendvai:
```
    (values
     (list*
      ;; NOTE: the order of the list elements influences various other
      ;; places in this file that call OUTPUT-FILE and take the NTH
      ;; member of the return value.
      (asdf:apply-output-translations generated-lisp-file)
      ;; If we unconditionally returned the SPEC-FILE here as an output
      ;; file, that would make ASDF unconditionally redo the rest of the
      ;; generation and compilation.
      (unless (probe-file spec-file)
        (list spec-file)))
     ;; Tell ASDF not to apply output translation.
     t)
```

src/enum.lisp
attila-lendvai: done

src/enum.lisp
attila-lendvai: done

src/enum.lisp
attila-lendvai: as an alternative, how about reformatting the rest of the loops?

i even volunteer doing that myself... because the non-keyword-colored loops are just plain unreadable for me.

but again, no attachment, i can live without making git blame useless... :)

src/enum.lisp
attila-lendvai: yeah, that's some unholy mixture of a thinko and some artifacts of the previous code. got rid of it altogether in a rebased version...

thanks!

src/types.lisp
attila-lendvai: the user-level issue is stuff like this:

```
(defcstruct linked-list
  (next (:struct linked-list)))
```

before the patch self-referential type lookup fails while defining the struct...

a simple fix is to introduce a stub for the struct type, and in case any errors happen, then just delete that stub from an UWP.

tests/bindings.lisp
attila-lendvai: yep, got rid of it.

tests/enum.lisp
attila-lendvai: because :long and :long-long is the same length on many paltforms. but then the same applies to :int and :long, so i rewrote this test.

tests/fsbv.lisp
attila-lendvai: hrm... i wrongly assumed at one point that enum's in C are always signed.

got rid of this change and recorded a commit to clean up the enum base type mapping.

attila-lendvai:
what i meant is e.g. that defcenum now expands to a defconstant for each member (unless it's a keyword).

no, i haven't seen the 32bit test breakage, got too much used to sbcl32 breaking. thanks for the heads up! should be fixed, it was due to my bogus assumptions on C type sizes in the test and enum code.

shall i push -f (inline comments will be gone), or close this PR and open a new one with the rebased commits?

attila-lendvai:
pushed a state that doesn't contain any pending issues that i know of.

re ppcre dependency: i added a slot to let people turn the default :cl-ppcre rule matcher into an :ends-with-subseq and thus don't require the full :cffi/c2ffi+cl-ppcre dependency (for an example see https://github.com/attila-lendvai/hu.dwim.bluez).

one questionable aspect of the current state: the matcher defaults to :cl-ppcre, which means that people need to depend on :cffi/c2ffi+cl-ppcre, not just :cffi/c2ffi. i was considering adding a :cffi/c2ffi-sans-cl-ppcre system, but that sounds even weirder.

sionescu:
Please just make cl-ppcre a default dependency.

attila-lendvai:
ok, i'll update the PR with a new commit that depends on cl-ppcre, so that it'll remain in the history at least if someone needs it.

background: one of my (probably cancelled) projects was targeted on a mips embedded device, and throwing in a cl-ppcre dependency is not irrelevant in such contexts.

luismbo:
Can we move the ASDF integration to a separate system that would only load cffi/c2ffi if the bindings are not yet generated?

luismbo:
And, come to think about it, that "seperate" system could simply be `:cffi`.

attila-lendvai:
that's a good idea @luismbo i'll incorporate that, too. but i think it's better to keep both of those systems separate from cffi, if for nothing else to serve as documentation for people first approaching the codebase.

luismbo:
@attila-lendvai ok, then perhaps place the asdf integration into `:cffi/c2ffi` and the rest into `:cffi/c2ffi-generator` or something. Does that make sense?

attila-lendvai:
i had exactly that on mind, naming included... :)

attila-lendvai:
purged my TODOs. i'm again not aware of any issues/requests, all pushed.

note: without this (few days old) ASDF fix https://github.com/fare/asdf/commit/4c71caad95e5ba44af75d2bfdb7fc15e80f05647 it may run the lisp file generation more often than necessary.

luismbo:
Great. Many thanks, Attila. The only thing missing is a manual chapter and at least a smoke test. As a maintainer, I think it'll actually be a good exercise for me to tackle those. :-) Please poke me if I take too long to do it.

Merged.

Symbolics:
> Great. Many thanks, Attila. The only thing missing is a manual chapter and at least a smoke test. As a maintainer, I think it'll actually be a good exercise for me to tackle those. :-) Please poke me if I take too long to do it.

Poke!

luismbo:
Ouch. 😅 I don't think it's going to happen this year... 😔

attila-lendvai:
@Symbolics FYI, i have pushed a few words into the README. it now contains a list of examples/projects that use cffi/c2ffi.

it's not much, but much more than nothing.

PR #86, opened 2016-03-14
Added expand-into-foreign-memory

cbaggers:
Very basic implementation of the feature. Almost a copy-paste of `expand-from-foreign`. Is used from `#'mem-set`.

I have tested with my projects and get the desired expansion.

I haven't added any tests to the suite as I don't see any tests the output of the current `expand-*` functions. If this is wrong please tell me how you'd like them written and I'll get that done.

### DIFF
--- a/src/structures.lisp
+++ b/src/structures.lisp
@@ -38,13 +38,6 @@
                    (let ((slot (gethash name (structure-slots type))))
                      (convert-to-foreign value (slot-type slot)))))))
 
-(defun convert-into-foreign-memory (value type ptr)
-  (let ((ptype (parse-type type)))
-    (if (typep ptype 'foreign-built-in-type)
-        value
-        (translate-into-foreign-memory value ptype ptr)))
-  ptr)
-
 (defmethod translate-to-foreign (value (type foreign-struct-type))
   (let ((ptr (foreign-alloc type)))
     (translate-into-foreign-memory value type ptr)
@@ -138,5 +131,3 @@
        (with-foreign-slots (,slot-names ,value-symbol ,struct-name)
          ,from-form))))
 |#
-
-
--- a/doc/cffi-manual.texinfo
+++ b/doc/cffi-manual.texinfo
@@ -148,23 +148,23 @@ software or the use or other dealings in the software.}
 
 @menu
 * Introduction::                What is CFFI?
-* Installation::                
-* Implementation Support::      
+* Installation::
+* Implementation Support::
 * Tutorial::                    Interactive intro to using CFFI.
 * Wrapper generators::          CFFI forms from munging C source code.
-* Foreign Types::               
-* Pointers::                    
-* Strings::                     
-* Variables::                   
-* Functions::                   
-* Libraries::                   
-* Callbacks::                   
-* The Groveller::               
+* Foreign Types::
+* Pointers::
+* Strings::
+* Variables::
+* Functions::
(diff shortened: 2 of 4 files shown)


luismbo:
Looks good. Should we export the symbol? Also, do you think `{translate,convert,expand}-into-foreign` (without the "memory" suffix) would be a better name?

cbaggers:
- The symbol was already exported.
- I'm undecided on the name issue. I like `into-memory` as it is different enough to be noticeable, and that is good given how the behavior is different from the others. However it is less consistent. I think I'd vote for keeping the current naming for the sake of backwards compatibility.

Hmm seems travis has an issue. And looks like something I was rather sure I had fixed already. I'll check that out

cbaggers:
Current travis error in sbcl32 step seems to be download related rather than this pr.

luismbo:
@cbaggers yes, I will disable sbcl32 to reduce the noise. Sorry about that.

cbaggers:
A quick 'best practices' question. Would you want me to squash these commits together in future?

luismbo:
Yes, definitely squash them at some point. I don't know what the best time is for doing that, but, yes, squashing will occur! :-)

cbaggers:
Squashed and ready to go

guicho271828:
documentation?

cbaggers:
@guicho271828 good point. I had forgotten that. I'll get on it

cbaggers:
Documentation written. Sorry for the delay.
I also noticed some of the documentation around `defcstruct` is a bit outdated so I'll update that soon. That will be a separate pull request. 
From my side I think this feature is complete now, but I'm happy to do more tweaking if needed.

cbaggers:
Ah I may have added a bug, may need some advice on this one.
I added a `convert-into-foreign-memory` definition (and matching compiler macro). but there is already one in `structures.lisp`

I guess I can just remove my version of the function but what behaviour do we want for the compiler macro?

luismbo:
Oh, you should remove the one in structures.lisp!

src/structures.lisp
```diff
@@ -38,13 +38,6 @@
                    (let ((slot (gethash name (structure-slots type))))
                      (convert-to-foreign value (slot-type slot)))))))
 
-(defun convert-into-foreign-memory (value type ptr)
-  (let ((ptype (parse-type type)))
-    (if (typep ptype 'foreign-built-in-type)
-        value
-        (translate-into-foreign-memory value ptype ptr)))
-  ptr)
```
luismbo: So, we have to be careful here because the definition of `convert-into-foreign-memory` has changed. This version looks wrong to me, but it is this way for some reason and code will break when we change it. We can either tackle that issue now or keep this strange version (but move it to `early-types.lisp`) and deal with that later. What do you feel like doing?

cbaggers:
Hmm, I think if we had found the bug in `convert-into-foreign-memory` outside of a feature branch we would just fix it as, like you say, it looks wrong. 

Given that I think that we should remove the old version and not just move it. The new version will actually call the  convert*` method for `foreign-built-in-type` too so that got to be better than just returning value :)

cbaggers:
Hey folks, is there anything I can do to help this process along?

luismbo:
You just did. :-) Thanks! (And sorry for the delay, this totally fell off my radar.)

cbaggers:
woohoo!

PR #88, opened 2016-04-09
Remove duplicate entries from *cc-flags*

mirkov:
From commit message:

`define-grovel-syntax pkg-config-cflags' would add the same paths even
if they were already in *cc-flags*.

The modified code appends the new flags, and then removes the
duplicates.  The new flags are placed at the end of *cc-flags*

### DIFF
--- a/grovel/grovel.lisp
+++ b/grovel/grovel.lisp
@@ -281,8 +281,13 @@ int main(int argc, char**argv) {
           (run-program program+args
                        :output (make-broadcast-stream output-stream *debug-io*)
                        :error-output output-stream)
-          (appendf *cc-flags*
-                   (parse-command-flags (get-output-stream-string output-stream))))
+          (setf *cc-flags* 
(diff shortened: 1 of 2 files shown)


luismbo:
@mirkov why are the two nreverse calls needed? Also, does this break if we have flags of the form "--flag <argument>"? (Are those possible?) Finally, is this fix simply cosmetic?

mirkov:
Two reverses: this was due to (possibly premature) optimization:  I wanted to append the flag to the existing list, not at the front.

Fix cosmetic: yes.  But also important not to confuse users that are trying to debug a compilation failure.  If a novice user sees a flag repeated, he/she may wonder about code correctness.

I have not checked the --flag form.  I do not recall pkg-config using that form.

sionescu:
Build tools should never hide user mistakes. Even if a flag is duplicate, if that is an error it should be corrected by the user.

PR #89, opened 2016-04-09
parse-command-flags: Added #\Return to separators

mirkov:
On Windows7+MSYS, the output of pkg-config as read by CCL included
\#RETURN.

I added #\Return the bag of separators.  I also reformatted the code a
bit, to make it a bit more readable (fewer long lines or ackward line
splits).


luismbo:
Thanks!

PR #90, opened 2016-05-18
use lisp-namespace for enabling documentation

guicho271828:
This makes define-foreign-library to accept (:documentation "str"). May require the next quicklisp release.

### DIFF
--- a/cffi.asd
+++ b/cffi.asd
@@ -36,7 +36,8 @@
   :author "James Bielman  <jamesjb@jamesjb.com>"
   :maintainer "Luis Oliveira  <loliveira@common-lisp.net>"
   :licence "MIT"
-  :depends-on (:uiop :alexandria :trivial-features :babel)
+  :depends-on (:uiop :alexandria :trivial-features :babel
+                     :lisp-namespace)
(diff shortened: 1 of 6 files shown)


luismbo:
`lisp-namespace` is a cool and useful idea. I'm curious about how an implementation/API based on CLOS would look like. At work we have a sort of `interned-object` class that'll cause objects of that class to have an associated namespace. Have you considered that approach?

Anyway, CFFI has so many dependant libraries that I'm becoming more cautious about taking on new (and super fresh) dependencies.

I've starred and subscribed to `lisp-namespace`, so I'll be keeping this in mind!

Thanks!

guicho271828:
I expected it will not be merged but I'm satisfied. Yes, it is obvious that cffi is one of the fundamental library. Also it is rare that someone C-c C-d the name of a foreign library.

Interned-object (an object associated with a symbol?) is that it will not solve the problem of "getting the documentation of a symbol immediately". Documentation is tied to an identifier (symbol) and not to the object.

luismbo:
Well, you can update the symbol documentation when the named object is instantiated, right?

PR #91, opened 2016-06-04
Respect CFLAGS environment variable

jasom:
We respect CC, CFLAGS seems useful as well.

I have some headers in non-standard locations, and this was the easiest way for me to access those.


jasom:
If you would rather I use split-sequence instead of a custom string splitting function, let me know.  It would add an additional dependency.

jasom:
Okay, the tests on travis revealed that my first attempt was stupid (It passed on my machine).  I also discovered CFFI already has a splitting function.

pipping:
This might be the right time to address a few bigger issues:
1. What is `*cc-flags*` anyway? It's used in `invoke-cc` which in turn is used in `cc-compile` and `link-executable`/`link-shared-library`. So if we wanted to pass anything to the preprocessor, the compiler, or the linker, we'd have to do it here. We should really be able to pass `CFLAGS`, `CPPFLAGS` (which is where your includes should rather go), and `LDFLAGS` but they shouldn't all go into just one big `*cc-flags*`.
2. `CC` and `CFLAGS` can be set by the user. But when and how often? Let me illustrate what I mean through a bit of sample code that's a reduction of what you propose (it'll only work for sbcl this way but could easily be adjusted):
```common-lisp
(defun getenv (var)
  ;; only implemented for sbcl here
  #+sbcl (sb-ext:posix-getenv var))

(defun setenv (var val)
  ;; only implemented for sbcl here
  #+sbcl (progn
           (require :sb-posix)
           (funcall (find-symbol (symbol-name :setenv) :sb-posix)
                    var val 1)))

(defparameter *cc-flags* (getenv "CFLAGS"))
(defparameter *cc* "cc")

(defun doit ()
  (format t "This is what we'd use: ~a ~a~%"
          (or (getenv "CC") *cc*)
          *cc-flags*))

(doit)
(setenv "CC" "cc1")
(setenv "CFLAGS" "cflags1")
(doit)
```
If I save that as foo.lisp and run
```shell
CC=cc0 CFLAGS=cflags0 sbcl --eval '(load "foo.lisp")' --quit
```
I get
```
This is what we'd use: cc0 cflags0
This is what we'd use: cc1 cflags0
```
because `(getenv "CC")` is called again for every `invoke-cc` whereas `(getenv "CFLAGS")` is only called when `defparameter` is evaluated. I'm not sure which behaviour is preferable but I think the choice should be consistent.

jasom:
If I could design how all of these worked it would be roughly:

    (defparameter *cflags* (or (getenv "CFLAGS") +default-cflags+))

Then if someone wants to override the cflags they can do:

    (let ((*cflags* ...)) ...)

That is to say: overriding CFFI's behavior from within the lisp image by setting environment variables is silly since dynamic bindings are so much more natural, and I see it unlikely that foreign code will invoke the groveller.

pipping:
@jasom I agree, setting environment variables from within lisp to affect the `CFLAGS` seems odd. That means we can move the `(or (getenv) ...)` logic for `CC` from `invoke-cc` to `*cc*` and have it run only once.

dkochmanski:
ECL has `compiler:*CC*`, `compiler:*USER-CC-FLAGS*` and `compiler:*USER-LD-FLAGS` plus a dozen of non-exported like `*ld*`, `*ld-flags*`, `*ld-rpath*` etc.., which defines are set during the compilation (queried from the environment etc.), so the custom flags are passed as dynamic variables.

luismbo:
@pipping's proposal makes perfect sense. @jasom, feel like adapting this PR along those lines?

luismbo:
BTW, it seems like a better idea to grab CFLAGS and friends at run time rather than load time. It will be much less surprising in the (perhaps unlikely) event that someone changes those variables from within the Lisp image and it'll play better with core dumping.

jasom:
That's a good point about saved images. I'll update the PR tonight (Los Angeles time).

jasom:
So what I do now is to load environment variables at both run-time and load-time.  If the dynamic variable has changed in any way (i.e been set or bound to something not EQ than it's load-time value), then it will ignore the environment variable, assuming the programmer knows what they are doing.

The only downside to what I have done is that users wanting to add options to the effective CFLAGS will have to use the function (cc-flags) to read the flags if they want to pickup on the new environment variable (I already changed the groveler to do this).  I see this as strictly better than other solutions though as the worst that will happen is they read from the specail variable and get whatever CFLAGS was at load-time.  More comments are in the new commit.

jasom:
ping

pipping:
I've just noticed that libffi/libffi-types.lisp calls a function `cc-flags` which I don't see defined anywhere and appears to pass the same includes (like e.g. -I/opt/local/include/ on darwin) as is already done here. That's... odd.

toolchain/c-toolchain.lisp
```diff
@@ -57,18 +57,25 @@
...
 
-(defparameter *cc-flags*
-  (append
-   ;; For MacPorts
-   #+darwin (list "-I" "/opt/local/include/")
-   ;; ECL internal flags
-   #+ecl (parse-command-flags c::*cc-flags*)
-   ;; FreeBSD non-base header files
-   #+freebsd (list "-I" "/usr/local/include/")))
+(defparameter *cc* *default-cc*)
```
luismbo: How about setting this to `:default` and testing that in `cc`? (Likewise for `cc-flags`.)

luismbo:
@pipping That's not a function call, it's a grovel directive.

toolchain/c-toolchain.lisp
jasom: I didn't do this because that changes the externally visible behavior (i.e. I don't know that someone out there isn't reading from *cc* or *cc-flags* to see what the default parameters are)

luismbo: I see.I suggest you don't worry about that, because the c-toolchain system is quite new and the only documented way of interacting with these was through the grovel directives. 

It's just that storing environment info from build time seems random and potentially confusing. (You've described this behaviour before, but I didn't fully grasp it then.) 

What do you think?

guicho271828:
ping

PR #92, opened 2016-06-11
Avoid using uoip:os-cond.

stassats:
It's only present in newer ASDFs and is basically a COND with
macroexpansion-time test evaluation. The code that uses it is not
performance critical.


pipping:
Since this issue continues to affect and confuse people (e.g. in #lisp on freenode), it seems like a good idea to make a release so that quicklisp can pick up this change.

luismbo:
@pipping done, thanks!

PR #93, opened 2016-06-21
ecl: defcallback: fix calling-convention

dkochmanski:
This is for bytecmp and the C compiler compatibility. :cdecl doesn't
work on Android, but :default works just fine. For C compiled code there
is no difference.

This patch will trigger itself on the upcoming ECL release (16.1.3) and is backward-compatible. Tested with cffi-tests – no regressions found (both with "fake" 16.1.3 version of ECL and 16.1.2).


luismbo:
Thanks!

PR #94, opened 2016-07-24
Small fixes and clarification

dfigrish:
I've spent a bit of time of my life to make sure that load-foreign-library should be used for reloading also. So, I'm pretty sure that it must be explicitly documented.


guicho271828:
This commit is one year old. What is the regular release cycle of CFFI?

guicho271828:
ping

attila-lendvai:
merged, thanks!

PR #96, opened 2016-09-01
don't depend on uiop if in sbcl, because sbcl already loads it.

jimka2001:
I'm taking out the explicit dependency on :uiop for sbcl because sbcl already loads it anyway
and including it causes lots of uiop functions to get needlessly redefined polluting startup with
volumes of useless warning messages.

### DIFF
--- a/cffi.asd
+++ b/cffi.asd
@@ -36,7 +36,8 @@
   :author "James Bielman  <jamesjb@jamesjb.com>"
   :maintainer "Luis Oliveira  <loliveira@common-lisp.net>"
   :licence "MIT"
-  :depends-on (:uiop :alexandria :trivial-features :babel)
+  :depends-on ( #-sbcl :uiop
+               :alexandria :trivial-features :babel)
   :components
   ((:module "src"
     :serial t


attila-lendvai:
FWIW, i'm not happy about this change, because it compromises cleanness to get rid of some warnings.

plus i think if you don't have a newer version of ASDF in the path, then it won't reload UIOP and ASDF and thus won't cause warnings. and if you do have, then you probably want it to get upgraded to the newest version.

jimka2001:
So is there a better test to make to keep cffi from re-loading the code?  I.e., a feature check rather than a check for sbcl.

jimka2001:
Here are the messages I get without this patch.

```
WARNING: redefining UIOP/PACKAGE:FIND-PACKAGE* in DEFUN
WARNING: redefining UIOP/PACKAGE:FIND-SYMBOL* in DEFUN
WARNING: redefining UIOP/PACKAGE:SYMBOL-CALL in DEFUN
WARNING: redefining UIOP/PACKAGE:INTERN* in DEFUN
WARNING: redefining UIOP/PACKAGE:EXPORT* in DEFUN
WARNING: redefining UIOP/PACKAGE:IMPORT* in DEFUN
WARNING: redefining UIOP/PACKAGE:SHADOWING-IMPORT* in DEFUN
WARNING: redefining UIOP/PACKAGE:SHADOW* in DEFUN
WARNING: redefining UIOP/PACKAGE:MAKE-SYMBOL* in DEFUN
WARNING: redefining UIOP/PACKAGE:UNINTERN* in DEFUN
WARNING: redefining UIOP/PACKAGE:SYMBOL-SHADOWING-P in DEFUN
WARNING: redefining UIOP/PACKAGE:HOME-PACKAGE-P in DEFUN
WARNING: redefining UIOP/PACKAGE:SYMBOL-PACKAGE-NAME in DEFUN
WARNING: redefining UIOP/PACKAGE:STANDARD-COMMON-LISP-SYMBOL-P in DEFUN
WARNING: redefining UIOP/PACKAGE:REIFY-PACKAGE in DEFUN
WARNING: redefining UIOP/PACKAGE:UNREIFY-PACKAGE in DEFUN
WARNING: redefining UIOP/PACKAGE:REIFY-SYMBOL in DEFUN
WARNING: redefining UIOP/PACKAGE:UNREIFY-SYMBOL in DEFUN
WARNING: redefining UIOP/PACKAGE::RECORD-FISHY in DEFUN
WARNING: redefining UIOP/PACKAGE::WHEN-PACKAGE-FISHINESS in DEFMACRO
WARNING: redefining UIOP/PACKAGE::NOTE-PACKAGE-FISHINESS in DEFMACRO
WARNING: redefining UIOP/PACKAGE::SET-DUMMY-SYMBOL in DEFUN
WARNING: redefining UIOP/PACKAGE::MAKE-DUMMY-SYMBOL in DEFUN
WARNING: redefining UIOP/PACKAGE::DUMMY-SYMBOL in DEFUN
WARNING: redefining UIOP/PACKAGE::GET-DUMMY-SYMBOL in DEFUN
WARNING: redefining UIOP/PACKAGE:NUKE-SYMBOL-IN-PACKAGE in DEFUN
WARNING: redefining UIOP/PACKAGE:NUKE-SYMBOL in DEFUN
WARNING: redefining UIOP/PACKAGE:REHOME-SYMBOL in DEFUN
WARNING: redefining UIOP/PACKAGE:ENSURE-PACKAGE-UNUSED in DEFUN
WARNING: redefining UIOP/PACKAGE:DELETE-PACKAGE* in DEFUN
WARNING: redefining UIOP/PACKAGE:PACKAGE-NAMES in DEFUN
WARNING: redefining UIOP/PACKAGE:PACKAGES-FROM-NAMES in DEFUN
WARNING: redefining UIOP/PACKAGE:FRESH-PACKAGE-NAME in DEFUN
WARNING: redefining UIOP/PACKAGE:RENAME-PACKAGE-AWAY in DEFUN
WARNING: redefining UIOP/PACKAGE:PACKAGE-DEFINITION-FORM in DEFUN
WARNING: redefining UIOP/PACKAGE::ENSURE-SHADOWING-IMPORT in DEFUN
WARNING: redefining UIOP/PACKAGE::ENSURE-IMPORTED in DEFUN
WARNING: redefining UIOP/PACKAGE::ENSURE-IMPORT in DEFUN
WARNING: redefining UIOP/PACKAGE::ENSURE-INHERITED in DEFUN
WARNING: redefining UIOP/PACKAGE::ENSURE-MIX in DEFUN
WARNING: redefining UIOP/PACKAGE::RECYCLE-SYMBOL in DEFUN
WARNING: redefining UIOP/PACKAGE::SYMBOL-RECYCLED-P in DEFUN
WARNING: redefining UIOP/PACKAGE::ENSURE-SYMBOL in DEFUN
WARNING: redefining UIOP/PACKAGE::ENSURE-EXPORTED-TO-USER in DEFUN
WARNING: redefining UIOP/PACKAGE::ENSURE-EXPORTED in DEFUN
WARNING: redefining UIOP/PACKAGE::ENSURE-EXPORT in DEFUN
WARNING: redefining UIOP/PACKAGE:ENSURE-PACKAGE in DEFUN
WARNING: redefining UIOP/PACKAGE:PARSE-DEFINE-PACKAGE-FORM in DEFUN
WARNING: redefining UIOP/PACKAGE:DEFINE-PACKAGE in DEFMACRO
WARNING: redefining UIOP/COMMON-LISP::LOOP* in DEFMACRO
WARNING: redefining UIOP/COMMON-LISP::FROB-SUBSTRINGS in DEFUN
WARNING: redefining UIOP/COMMON-LISP::COMPATFMT in DEFMACRO
WARNING: redefining UIOP/UTILITY:UNDEFINE-FUNCTION in DEFUN
WARNING: redefining UIOP/UTILITY:UNDEFINE-FUNCTIONS in DEFUN
WARNING: redefining UIOP/UTILITY:DEFGENERIC* in DEFMACRO
WARNING: redefining UIOP/UTILITY:DEFUN* in DEFMACRO
WARNING: redefining UIOP/UTILITY:WITH-UPGRADABILITY in DEFMACRO
WARNING: redefining UIOP/UTILITY:UIOP-DEBUG in DEFMACRO
WARNING: redefining UIOP/UTILITY:LOAD-UIOP-DEBUG-UTILITY in DEFUN
WARNING: redefining UIOP/UTILITY:NEST in DEFMACRO
WARNING: redefining UIOP/UTILITY:IF-LET in DEFMACRO
WARNING: redefining UIOP/UTILITY:WHILE-COLLECTING in DEFMACRO
WARNING: redefining UIOP/UTILITY:APPENDF in DEFMACRO
WARNING: redefining UIOP/UTILITY:LENGTH=N-P in DEFUN
WARNING: redefining UIOP/UTILITY:ENSURE-LIST in DEFUN
WARNING: redefining UIOP/UTILITY:REMOVE-PLIST-KEY in DEFUN
WARNING: redefining UIOP/UTILITY:REMOVE-PLIST-KEYS in DEFUN
WARNING: redefining UIOP/UTILITY:EMPTYP in DEFUN
WARNING: redefining UIOP/UTILITY:CHARACTER-TYPE-INDEX in DEFUN
WARNING: redefining UIOP/UTILITY:BASE-STRING-P in DEFUN
WARNING: redefining UIOP/UTILITY:STRINGS-COMMON-ELEMENT-TYPE in DEFUN
WARNING: redefining UIOP/UTILITY:REDUCE/STRCAT in DEFUN
WARNING: redefining UIOP/UTILITY:STRCAT in DEFUN
WARNING: redefining UIOP/UTILITY:FIRST-CHAR in DEFUN
WARNING: redefining UIOP/UTILITY:LAST-CHAR in DEFUN
WARNING: redefining UIOP/UTILITY:SPLIT-STRING in DEFUN
WARNING: redefining UIOP/UTILITY:STRING-PREFIX-P in DEFUN
WARNING: redefining UIOP/UTILITY:STRING-SUFFIX-P in DEFUN
WARNING: redefining UIOP/UTILITY:STRING-ENCLOSED-P in DEFUN
WARNING: redefining UIOP/UTILITY:STRIPLN in DEFUN
WARNING: redefining UIOP/UTILITY:STANDARD-CASE-SYMBOL-NAME in DEFUN
WARNING: redefining UIOP/UTILITY:FIND-STANDARD-CASE-SYMBOL in DEFUN
WARNING: redefining UIOP/UTILITY:STAMP< in DEFUN
WARNING: redefining UIOP/UTILITY:STAMPS< in DEFUN
WARNING: redefining UIOP/UTILITY:STAMP*< in DEFUN
WARNING: redefining UIOP/UTILITY:STAMP<= in DEFUN
WARNING: redefining UIOP/UTILITY:EARLIER-STAMP in DEFUN
WARNING: redefining UIOP/UTILITY:STAMPS-EARLIEST in DEFUN
WARNING: redefining UIOP/UTILITY:EARLIEST-STAMP in DEFUN
WARNING: redefining UIOP/UTILITY:LATER-STAMP in DEFUN
WARNING: redefining UIOP/UTILITY:STAMPS-LATEST in DEFUN
WARNING: redefining UIOP/UTILITY:LATEST-STAMP in DEFUN
WARNING: redefining UIOP/UTILITY:LATEST-STAMP-F in DEFMACRO
WARNING: redefining UIOP/UTILITY:ENSURE-FUNCTION in DEFUN
WARNING: redefining UIOP/UTILITY:ACCESS-AT in DEFUN
WARNING: redefining UIOP/UTILITY:ACCESS-AT-COUNT in DEFUN
WARNING: redefining UIOP/UTILITY:CALL-FUNCTION in DEFUN
WARNING: redefining UIOP/UTILITY:CALL-FUNCTIONS in DEFUN
WARNING: redefining UIOP/UTILITY:REGISTER-HOOK-FUNCTION in DEFUN
WARNING: redefining UIOP/UTILITY:COERCE-CLASS in DEFUN
WARNING: redefining UIOP/UTILITY:ENSURE-GETHASH in DEFUN
WARNING: redefining UIOP/UTILITY:LIST-TO-HASH-SET in DEFUN
WARNING: redefining UIOP/UTILITY:UNPARSE-VERSION in DEFUN
WARNING: redefining UIOP/UTILITY:PARSE-VERSION in DEFUN
WARNING: redefining UIOP/UTILITY:LEXICOGRAPHIC< in DEFUN
WARNING: redefining UIOP/UTILITY:LEXICOGRAPHIC<= in DEFUN
WARNING: redefining UIOP/UTILITY:VERSION< in DEFUN
WARNING: redefining UIOP/UTILITY:VERSION<= in DEFUN
WARNING: redefining UIOP/UTILITY:VERSION-COMPATIBLE-P in DEFUN
WARNING: redefining UIOP/UTILITY:MATCH-CONDITION-P in DEFUN
WARNING: redefining UIOP/UTILITY:MATCH-ANY-CONDITION-P in DEFUN
WARNING: redefining UIOP/UTILITY:CALL-WITH-MUFFLED-CONDITIONS in DEFUN
WARNING: redefining UIOP/UTILITY:WITH-MUFFLED-CONDITIONS in DEFMACRO
WARNING: redefining UIOP/OS:FEATUREP in DEFUN
WARNING: redefining UIOP/OS:OS-MACOSX-P in DEFUN
WARNING: redefining UIOP/OS:OS-UNIX-P in DEFUN
WARNING: redefining UIOP/OS:OS-WINDOWS-P in DEFUN
WARNING: redefining UIOP/OS:OS-GENERA-P in DEFUN
WARNING: redefining UIOP/OS::OS-OLDMAC-P in DEFUN
WARNING: redefining UIOP/OS:DETECT-OS in DEFUN
WARNING: redefining UIOP/OS:OS-COND in DEFMACRO
WARNING: redefining UIOP/OS:GETENV in DEFUN
WARNING: redefining UIOP/OS:GETENVP in DEFUN
WARNING: redefining UIOP/OS::FIRST-FEATURE in DEFUN
WARNING: redefining UIOP/OS:IMPLEMENTATION-TYPE in DEFUN
WARNING: redefining UIOP/OS:OPERATING-SYSTEM in DEFUN
WARNING: redefining UIOP/OS:ARCHITECTURE in DEFUN
WARNING: redefining UIOP/OS:LISP-VERSION-STRING in DEFUN
WARNING: redefining UIOP/OS:IMPLEMENTATION-IDENTIFIER in DEFUN
WARNING: redefining UIOP/OS:HOSTNAME in DEFUN
WARNING: redefining UIOP/OS:GETCWD in DEFUN
WARNING: redefining UIOP/OS:CHDIR in DEFUN
WARNING: redefining UIOP/OS:READ-NULL-TERMINATED-STRING in DEFUN
WARNING: redefining UIOP/OS:READ-LITTLE-ENDIAN in DEFUN
WARNING: redefining UIOP/OS:PARSE-FILE-LOCATION-INFO in DEFUN
WARNING: redefining UIOP/OS:PARSE-WINDOWS-SHORTCUT in DEFUN
WARNING:
   redefining UIOP/PATHNAME:NORMALIZE-PATHNAME-DIRECTORY-COMPONENT in DEFUN
WARNING:
   redefining UIOP/PATHNAME:DENORMALIZE-PATHNAME-DIRECTORY-COMPONENT in DEFUN
WARNING: redefining UIOP/PATHNAME:MERGE-PATHNAME-DIRECTORY-COMPONENTS in DEFUN
WARNING: redefining UIOP/PATHNAME:MAKE-PATHNAME* in DEFUN
WARNING: redefining UIOP/PATHNAME:MAKE-PATHNAME-COMPONENT-LOGICAL in DEFUN
WARNING: redefining UIOP/PATHNAME:MAKE-PATHNAME-LOGICAL in DEFUN
WARNING: redefining UIOP/PATHNAME:MERGE-PATHNAMES* in DEFUN
WARNING: redefining UIOP/PATHNAME:LOGICAL-PATHNAME-P in DEFUN
WARNING: redefining UIOP/PATHNAME:PHYSICAL-PATHNAME-P in DEFUN
WARNING: redefining UIOP/PATHNAME:PHYSICALIZE-PATHNAME in DEFUN
WARNING: redefining UIOP/PATHNAME:NIL-PATHNAME in DEFUN
WARNING: redefining UIOP/PATHNAME:WITH-PATHNAME-DEFAULTS in DEFMACRO
WARNING: redefining UIOP/PATHNAME:PATHNAME-EQUAL in DEFUN
WARNING: redefining UIOP/PATHNAME:ABSOLUTE-PATHNAME-P in DEFUN
WARNING: redefining UIOP/PATHNAME:RELATIVE-PATHNAME-P in DEFUN
WARNING: redefining UIOP/PATHNAME:HIDDEN-PATHNAME-P in DEFUN
WARNING: redefining UIOP/PATHNAME:FILE-PATHNAME-P in DEFUN
WARNING: redefining UIOP/PATHNAME:PATHNAME-DIRECTORY-PATHNAME in DEFUN
WARNING: redefining UIOP/PATHNAME:PATHNAME-PARENT-DIRECTORY-PATHNAME in DEFUN
WARNING: redefining UIOP/PATHNAME:DIRECTORY-PATHNAME-P in DEFUN
WARNING: redefining UIOP/PATHNAME:ENSURE-DIRECTORY-PATHNAME in DEFUN
WARNING:
   redefining UIOP/PATHNAME:SPLIT-UNIX-NAMESTRING-DIRECTORY-COMPONENTS in DEFUN
WARNING: redefining UIOP/PATHNAME:SPLIT-NAME-TYPE in DEFUN
WARNING: redefining UIOP/PATHNAME:PARSE-UNIX-NAMESTRING in DEFUN
WARNING: redefining UIOP/PATHNAME:UNIX-NAMESTRING in DEFUN
WARNING: redefining UIOP/PATHNAME:SUBPATHNAME in DEFUN
WARNING: redefining UIOP/PATHNAME:SUBPATHNAME* in DEFUN
WARNING: redefining UIOP/PATHNAME:PATHNAME-ROOT in DEFUN
WARNING: redefining UIOP/PATHNAME:PATHNAME-HOST-PATHNAME in DEFUN
WARNING: redefining UIOP/PATHNAME:ENSURE-ABSOLUTE-PATHNAME in DEFUN
WARNING: redefining UIOP/PATHNAME:SUBPATHP in DEFUN
WARNING: redefining UIOP/PATHNAME:ENOUGH-PATHNAME in DEFUN
WARNING: redefining UIOP/PATHNAME:CALL-WITH-ENOUGH-PATHNAME in DEFUN
WARNING: redefining UIOP/PATHNAME:WITH-ENOUGH-PATHNAME in DEFMACRO
WARNING: redefining UIOP/PATHNAME:WILDEN in DEFUN
WARNING: redefining UIOP/PATHNAME:RELATIVIZE-DIRECTORY-COMPONENT in DEFUN
WARNING: redefining UIOP/PATHNAME:RELATIVIZE-PATHNAME-DIRECTORY in DEFUN
WARNING: redefining UIOP/PATHNAME:DIRECTORY-SEPARATOR-FOR-HOST in DEFUN
WARNING: redefining UIOP/PATHNAME:DIRECTORIZE-PATHNAME-HOST-DEVICE in DEFUN
WARNING: redefining UIOP/FILESYSTEM:NATIVE-NAMESTRING in DEFUN
WARNING: redefining UIOP/FILESYSTEM:PARSE-NATIVE-NAMESTRING in DEFUN
WARNING: redefining UIOP/FILESYSTEM:TRUENAME* in DEFUN
WARNING: redefining UIOP/FILESYSTEM:SAFE-FILE-WRITE-DATE in DEFUN
WARNING: redefining UIOP/FILESYSTEM:PROBE-FILE* in DEFUN
WARNING: redefining UIOP/FILESYSTEM:DIRECTORY-EXISTS-P in DEFUN
WARNING: redefining UIOP/FILESYSTEM:FILE-EXISTS-P in DEFUN
WARNING: redefining UIOP/FILESYSTEM:DIRECTORY* in DEFUN
WARNING: redefining UIOP/FILESYSTEM:FILTER-LOGICAL-DIRECTORY-RESULTS in DEFUN
WARNING: redefining UIOP/FILESYSTEM:DIRECTORY-FILES in DEFUN
WARNING: redefining UIOP/FILESYSTEM:SUBDIRECTORIES in DEFUN
WARNING: redefining UIOP/FILESYSTEM:COLLECT-SUB*DIRECTORIES in DEFUN
WARNING: redefining UIOP/FILESYSTEM:TRUENAMIZE in DEFUN
WARNING: redefining UIOP/FILESYSTEM:RESOLVE-SYMLINKS in DEFUN
WARNING: redefining UIOP/FILESYSTEM:RESOLVE-SYMLINKS* in DEFUN
WARNING: redefining UIOP/PATHNAME:ENSURE-PATHNAME in DEFUN
WARNING: redefining UIOP/FILESYSTEM:GET-PATHNAME-DEFAULTS in DEFUN
WARNING: redefining UIOP/FILESYSTEM:CALL-WITH-CURRENT-DIRECTORY in DEFUN
WARNING: redefining UIOP/FILESYSTEM:WITH-CURRENT-DIRECTORY in DEFMACRO
WARNING: redefining UIOP/FILESYSTEM:INTER-DIRECTORY-SEPARATOR in DEFUN
WARNING: redefining UIOP/FILESYSTEM:SPLIT-NATIVE-PATHNAMES-STRING in DEFUN
WARNING: redefining UIOP/FILESYSTEM:GETENV-PATHNAME in DEFUN
WARNING: redefining UIOP/FILESYSTEM:GETENV-PATHNAMES in DEFUN
WARNING: redefining UIOP/FILESYSTEM:GETENV-ABSOLUTE-DIRECTORY in DEFUN
WARNING: redefining UIOP/FILESYSTEM:GETENV-ABSOLUTE-DIRECTORIES in DEFUN
WARNING: redefining UIOP/FILESYSTEM:LISP-IMPLEMENTATION-DIRECTORY in DEFUN
WARNING: redefining UIOP/FILESYSTEM:LISP-IMPLEMENTATION-PATHNAME-P in DEFUN
WARNING: redefining UIOP/FILESYSTEM:ENSURE-ALL-DIRECTORIES-EXIST in DEFUN
WARNING: redefining UIOP/FILESYSTEM:DELETE-FILE-IF-EXISTS in DEFUN
WARNING: redefining UIOP/FILESYSTEM:RENAME-FILE-OVERWRITING-TARGET in DEFUN
WARNING: redefining UIOP/FILESYSTEM:DELETE-EMPTY-DIRECTORY in DEFUN
WARNING: redefining UIOP/FILESYSTEM:DELETE-DIRECTORY-TREE in DEFUN
WARNING: redefining UIOP/STREAM:SETUP-STDIN in DEFUN
WARNING: redefining UIOP/STREAM:SETUP-STDOUT in DEFUN
WARNING: redefining UIOP/STREAM:SETUP-STDERR in DEFUN
WARNING: redefining UIOP/STREAM:ALWAYS-DEFAULT-ENCODING in DEFUN
WARNING: redefining UIOP/STREAM:DETECT-ENCODING in DEFUN
WARNING: redefining UIOP/STREAM:DEFAULT-ENCODING-EXTERNAL-FORMAT in DEFUN
WARNING: redefining UIOP/STREAM:ENCODING-EXTERNAL-FORMAT in DEFUN
WARNING: redefining UIOP/STREAM:WITH-SAFE-IO-SYNTAX in DEFMACRO
WARNING: redefining UIOP/STREAM:CALL-WITH-SAFE-IO-SYNTAX in DEFUN
WARNING: redefining UIOP/STREAM:SAFE-READ-FROM-STRING in DEFUN
WARNING: redefining UIOP/STREAM:CALL-WITH-OUTPUT-FILE in DEFUN
WARNING: redefining UIOP/STREAM:WITH-OUTPUT-FILE in DEFMACRO
WARNING: redefining UIOP/STREAM::CALL-WITH-OUTPUT in DEFUN
WARNING: redefining UIOP/STREAM:WITH-OUTPUT in DEFMACRO
WARNING: redefining UIOP/STREAM:OUTPUT-STRING in DEFUN
WARNING: redefining UIOP/STREAM:CALL-WITH-INPUT-FILE in DEFUN
WARNING: redefining UIOP/STREAM:WITH-INPUT-FILE in DEFMACRO
WARNING: redefining UIOP/STREAM::CALL-WITH-INPUT in DEFUN
WARNING: redefining UIOP/STREAM:WITH-INPUT in DEFMACRO
WARNING: redefining UIOP/STREAM:INPUT-STRING in DEFUN
WARNING: redefining UIOP/STREAM:NULL-DEVICE-PATHNAME in DEFUN
WARNING: redefining UIOP/STREAM:CALL-WITH-NULL-INPUT in DEFUN
WARNING: redefining UIOP/STREAM:WITH-NULL-INPUT in DEFMACRO
WARNING: redefining UIOP/STREAM:CALL-WITH-NULL-OUTPUT in DEFUN
WARNING: redefining UIOP/STREAM:WITH-NULL-OUTPUT in DEFMACRO
WARNING: redefining UIOP/STREAM:FINISH-OUTPUTS in DEFUN
WARNING: redefining UIOP/STREAM:FORMAT! in DEFUN
WARNING: redefining UIOP/STREAM:SAFE-FORMAT! in DEFUN
WARNING: redefining UIOP/STREAM:COPY-STREAM-TO-STREAM in DEFUN
WARNING: redefining UIOP/STREAM:CONCATENATE-FILES in DEFUN
WARNING: redefining UIOP/STREAM:COPY-FILE in DEFUN
WARNING: redefining UIOP/STREAM:SLURP-STREAM-STRING in DEFUN
WARNING: redefining UIOP/STREAM:SLURP-STREAM-LINES in DEFUN
WARNING: redefining UIOP/STREAM:SLURP-STREAM-LINE in DEFUN
WARNING: redefining UIOP/STREAM:SLURP-STREAM-FORMS in DEFUN
WARNING: redefining UIOP/STREAM:SLURP-STREAM-FORM in DEFUN
WARNING: redefining UIOP/STREAM:READ-FILE-STRING in DEFUN
WARNING: redefining UIOP/STREAM:READ-FILE-LINES in DEFUN
WARNING: redefining UIOP/STREAM:READ-FILE-LINE in DEFUN
WARNING: redefining UIOP/STREAM:READ-FILE-FORMS in DEFUN
WARNING: redefining UIOP/STREAM:READ-FILE-FORM in DEFUN
WARNING: redefining UIOP/STREAM:SAFE-READ-FILE-LINE in DEFUN
WARNING: redefining UIOP/STREAM:SAFE-READ-FILE-FORM in DEFUN
WARNING: redefining UIOP/STREAM:EVAL-INPUT in DEFUN
WARNING: redefining UIOP/STREAM:EVAL-THUNK in DEFUN
WARNING: redefining UIOP/STREAM:STANDARD-EVAL-THUNK in DEFUN
WARNING: redefining UIOP/STREAM:PRINTLN in DEFUN
WARNING: redefining UIOP/STREAM:WRITELN in DEFUN
WARNING: redefining UIOP/STREAM:DEFAULT-TEMPORARY-DIRECTORY in DEFUN
WARNING: redefining UIOP/STREAM:TEMPORARY-DIRECTORY in DEFUN
WARNING: redefining UIOP/STREAM:SETUP-TEMPORARY-DIRECTORY in DEFUN
WARNING: redefining UIOP/STREAM:CALL-WITH-TEMPORARY-FILE in DEFUN
WARNING: redefining UIOP/STREAM:WITH-TEMPORARY-FILE in DEFMACRO
WARNING: redefining UIOP/STREAM::GET-TEMPORARY-FILE in DEFUN
WARNING: redefining UIOP/STREAM:ADD-PATHNAME-SUFFIX in DEFUN
WARNING: redefining UIOP/STREAM:TMPIZE-PATHNAME in DEFUN
WARNING: redefining UIOP/STREAM:CALL-WITH-STAGING-PATHNAME in DEFUN
WARNING: redefining UIOP/STREAM:WITH-STAGING-PATHNAME in DEFMACRO
WARNING: redefining UIOP/IMAGE:QUIT in DEFUN
WARNING: redefining UIOP/IMAGE:DIE in DEFUN
WARNING: redefining UIOP/IMAGE:RAW-PRINT-BACKTRACE in DEFUN
WARNING: redefining UIOP/IMAGE:PRINT-BACKTRACE in DEFUN
WARNING: redefining UIOP/IMAGE:PRINT-CONDITION-BACKTRACE in DEFUN
WARNING: redefining UIOP/IMAGE:FATAL-CONDITION-P in DEFUN
WARNING: redefining UIOP/IMAGE:HANDLE-FATAL-CONDITION in DEFUN
WARNING: redefining UIOP/IMAGE:CALL-WITH-FATAL-CONDITION-HANDLER in DEFUN
WARNING: redefining UIOP/IMAGE:WITH-FATAL-CONDITION-HANDLER in DEFMACRO
WARNING: redefining UIOP/IMAGE:SHELL-BOOLEAN-EXIT in DEFUN
WARNING: redefining UIOP/IMAGE:REGISTER-IMAGE-RESTORE-HOOK in DEFUN
WARNING: redefining UIOP/IMAGE:REGISTER-IMAGE-DUMP-HOOK in DEFUN
WARNING: redefining UIOP/IMAGE:CALL-IMAGE-RESTORE-HOOK in DEFUN
WARNING: redefining UIOP/IMAGE:CALL-IMAGE-DUMP-HOOK in DEFUN
WARNING: redefining UIOP/IMAGE:RAW-COMMAND-LINE-ARGUMENTS in DEFUN
WARNING: redefining UIOP/IMAGE:COMMAND-LINE-ARGUMENTS in DEFUN
WARNING: redefining UIOP/IMAGE:ARGV0 in DEFUN
WARNING: redefining UIOP/IMAGE:SETUP-COMMAND-LINE-ARGUMENTS in DEFUN
WARNING: redefining UIOP/IMAGE:RESTORE-IMAGE in DEFUN
WARNING: redefining UIOP/IMAGE:DUMP-IMAGE in DEFUN
WARNING: redefining UIOP/IMAGE:CREATE-IMAGE in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::REQUIRES-ESCAPING-P in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM:ESCAPE-TOKEN in DEFUN
WARNING:
   redefining UIOP/RUN-PROGRAM::ESCAPE-WINDOWS-TOKEN-WITHIN-DOUBLE-QUOTES in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::EASY-WINDOWS-CHARACTER-P in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM:ESCAPE-WINDOWS-TOKEN in DEFUN
WARNING:
   redefining UIOP/RUN-PROGRAM::ESCAPE-SH-TOKEN-WITHIN-DOUBLE-QUOTES in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM:EASY-SH-CHARACTER-P in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM:ESCAPE-SH-TOKEN in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::ESCAPE-SHELL-TOKEN in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM:ESCAPE-COMMAND in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM:ESCAPE-WINDOWS-COMMAND in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM:ESCAPE-SH-COMMAND in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::ESCAPE-SHELL-COMMAND in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::CALL-STREAM-PROCESSOR in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM:SLURP-INPUT-STREAM in DEFGENERIC
WARNING:
   redefining SLURP-INPUT-STREAM (#<SB-PCL:SYSTEM-CLASS COMMON-LISP:FUNCTION>
                                  #<SB-PCL:SYSTEM-CLASS
                                    COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining SLURP-INPUT-STREAM (#<BUILT-IN-CLASS COMMON-LISP:CONS>
                                  #<SB-PCL:SYSTEM-CLASS
                                    COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining SLURP-INPUT-STREAM (#<SB-PCL:SYSTEM-CLASS COMMON-LISP:STREAM>
                                  #<SB-PCL:SYSTEM-CLASS
                                    COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining SLURP-INPUT-STREAM (#<SB-MOP:EQL-SPECIALIZER STRING>
                                  #<SB-PCL:SYSTEM-CLASS
                                    COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining SLURP-INPUT-STREAM (#<SB-MOP:EQL-SPECIALIZER :STRING>
                                  #<SB-PCL:SYSTEM-CLASS
                                    COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining SLURP-INPUT-STREAM (#<SB-MOP:EQL-SPECIALIZER :LINES>
                                  #<SB-PCL:SYSTEM-CLASS
                                    COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining SLURP-INPUT-STREAM (#<SB-MOP:EQL-SPECIALIZER :LINE>
                                  #<SB-PCL:SYSTEM-CLASS
                                    COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining SLURP-INPUT-STREAM (#<SB-MOP:EQL-SPECIALIZER :FORMS>
                                  #<SB-PCL:SYSTEM-CLASS
                                    COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining SLURP-INPUT-STREAM (#<SB-MOP:EQL-SPECIALIZER :FORM>
                                  #<SB-PCL:SYSTEM-CLASS
                                    COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining SLURP-INPUT-STREAM (#<SB-MOP:EQL-SPECIALIZER T>
                                  #<SB-PCL:SYSTEM-CLASS
                                    COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining SLURP-INPUT-STREAM (#<BUILT-IN-CLASS COMMON-LISP:NULL>
                                  #<SB-PCL:SYSTEM-CLASS
                                    COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining SLURP-INPUT-STREAM (#<STRUCTURE-CLASS COMMON-LISP:PATHNAME>
                                  #<SB-PCL:SYSTEM-CLASS
                                    COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining SLURP-INPUT-STREAM (#<SB-PCL:SYSTEM-CLASS COMMON-LISP:T>
                                  #<SB-PCL:SYSTEM-CLASS
                                    COMMON-LISP:T>) in DEFMETHOD
WARNING: redefining UIOP/RUN-PROGRAM:VOMIT-OUTPUT-STREAM in DEFGENERIC
WARNING:
   redefining VOMIT-OUTPUT-STREAM (#<SB-PCL:SYSTEM-CLASS COMMON-LISP:FUNCTION>
                                   #<SB-PCL:SYSTEM-CLASS
                                     COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining VOMIT-OUTPUT-STREAM (#<BUILT-IN-CLASS COMMON-LISP:CONS>
                                   #<SB-PCL:SYSTEM-CLASS
                                     COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining VOMIT-OUTPUT-STREAM (#<SB-PCL:SYSTEM-CLASS COMMON-LISP:STREAM>
                                   #<SB-PCL:SYSTEM-CLASS
                                     COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining VOMIT-OUTPUT-STREAM (#<BUILT-IN-CLASS COMMON-LISP:STRING>
                                   #<SB-PCL:SYSTEM-CLASS
                                     COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining VOMIT-OUTPUT-STREAM (#<SB-MOP:EQL-SPECIALIZER T>
                                   #<SB-PCL:SYSTEM-CLASS
                                     COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining VOMIT-OUTPUT-STREAM (#<BUILT-IN-CLASS COMMON-LISP:NULL>
                                   #<SB-PCL:SYSTEM-CLASS
                                     COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining VOMIT-OUTPUT-STREAM (#<STRUCTURE-CLASS COMMON-LISP:PATHNAME>
                                   #<SB-PCL:SYSTEM-CLASS
                                     COMMON-LISP:T>) in DEFMETHOD
WARNING:
   redefining VOMIT-OUTPUT-STREAM (#<SB-PCL:SYSTEM-CLASS COMMON-LISP:T>
                                   #<SB-PCL:SYSTEM-CLASS
                                     COMMON-LISP:T>) in DEFMETHOD
WARNING: redefining UIOP/RUN-PROGRAM::%CMD-SHELL-PATHNAME in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::%NORMALIZE-COMMAND in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::%ACTIVE-IO-SPECIFIER-P in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::%NORMALIZE-IO-SPECIFIER in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::%INTERACTIVEP in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::%RUN-PROGRAM in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::%PROCESS-INFO-PID in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::%WAIT-PROCESS-RESULT in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::%CHECK-RESULT in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::%CALL-WITH-PROGRAM-IO in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::PLACE-SETTER in DEFMACRO
WARNING: redefining UIOP/RUN-PROGRAM::WITH-PROGRAM-INPUT in DEFMACRO
WARNING: redefining UIOP/RUN-PROGRAM::WITH-PROGRAM-OUTPUT in DEFMACRO
WARNING: redefining UIOP/RUN-PROGRAM::WITH-PROGRAM-ERROR-OUTPUT in DEFMACRO
WARNING: redefining UIOP/RUN-PROGRAM::%USE-RUN-PROGRAM in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::%NORMALIZE-SYSTEM-COMMAND in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::%REDIRECTED-SYSTEM-COMMAND in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::%SYSTEM in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM::%USE-SYSTEM in DEFUN
WARNING: redefining UIOP/RUN-PROGRAM:RUN-PROGRAM in DEFUN
WARNING: redefining UIOP/LISP-BUILD:GET-OPTIMIZATION-SETTINGS in DEFUN
WARNING: redefining UIOP/LISP-BUILD:PROCLAIM-OPTIMIZATION-SETTINGS in DEFUN
WARNING: redefining UIOP/LISP-BUILD:WITH-OPTIMIZATION-SETTINGS in DEFMACRO
WARNING:
   redefining UIOP/LISP-BUILD::SB-GROVEL-UNKNOWN-CONSTANT-CONDITION-P in DEFUN
WARNING:
   redefining UIOP/LISP-BUILD:CALL-WITH-MUFFLED-COMPILER-CONDITIONS in DEFUN
WARNING: redefining UIOP/LISP-BUILD:WITH-MUFFLED-COMPILER-CONDITIONS in DEFMACRO
WARNING: redefining UIOP/LISP-BUILD:CALL-WITH-MUFFLED-LOADER-CONDITIONS in DEFUN
WARNING: redefining UIOP/LISP-BUILD:WITH-MUFFLED-LOADER-CONDITIONS in DEFMACRO
WARNING: redefining UIOP/LISP-BUILD:CHECK-LISP-COMPILE-WARNINGS in DEFUN
WARNING: redefining UIOP/LISP-BUILD:CHECK-LISP-COMPILE-RESULTS in DEFUN
WARNING: redefining UIOP/LISP-BUILD:REIFY-SIMPLE-SEXP in DEFUN
WARNING: redefining UIOP/LISP-BUILD:UNREIFY-SIMPLE-SEXP in DEFUN
WARNING: redefining UIOP/LISP-BUILD::REIFY-UNDEFINED-WARNING in DEFUN
WARNING: redefining UIOP/LISP-BUILD:REIFY-DEFERRED-WARNINGS in DEFUN
WARNING: redefining UIOP/LISP-BUILD:UNREIFY-DEFERRED-WARNINGS in DEFUN
WARNING: redefining UIOP/LISP-BUILD:RESET-DEFERRED-WARNINGS in DEFUN
WARNING: redefining UIOP/LISP-BUILD:SAVE-DEFERRED-WARNINGS in DEFUN
WARNING: redefining UIOP/LISP-BUILD:WARNINGS-FILE-TYPE in DEFUN
WARNING: redefining UIOP/LISP-BUILD:ENABLE-DEFERRED-WARNINGS-CHECK in DEFUN
WARNING: redefining UIOP/LISP-BUILD:DISABLE-DEFERRED-WARNINGS-CHECK in DEFUN
WARNING: redefining UIOP/LISP-BUILD:WARNINGS-FILE-P in DEFUN
WARNING: redefining UIOP/LISP-BUILD:CHECK-DEFERRED-WARNINGS in DEFUN
WARNING: redefining UIOP/LISP-BUILD::CALL-WITH-SAVED-DEFERRED-WARNINGS in DEFUN
WARNING: redefining UIOP/LISP-BUILD:WITH-SAVED-DEFERRED-WARNINGS in DEFMACRO
WARNING: redefining UIOP/LISP-BUILD:CURRENT-LISP-FILE-PATHNAME in DEFUN
WARNING: redefining UIOP/LISP-BUILD:LOAD-PATHNAME in DEFUN
WARNING: redefining UIOP/LISP-BUILD:LISPIZE-PATHNAME in DEFUN
WARNING: redefining UIOP/LISP-BUILD:COMPILE-FILE-TYPE in DEFUN
WARNING: redefining UIOP/LISP-BUILD:CALL-AROUND-HOOK in DEFUN
WARNING: redefining UIOP/LISP-BUILD:COMPILE-FILE-PATHNAME* in DEFUN
WARNING: redefining UIOP/LISP-BUILD:LOAD* in DEFUN
WARNING: redefining UIOP/LISP-BUILD:LOAD-FROM-STRING in DEFUN
WARNING: redefining UIOP/LISP-BUILD:COMBINE-FASLS in DEFUN
WARNING:
   redefining UIOP/CONFIGURATION:CONFIGURATION-INHERITANCE-DIRECTIVE-P in DEFUN
WARNING: redefining UIOP/CONFIGURATION:REPORT-INVALID-FORM in DEFUN
WARNING: redefining UIOP/CONFIGURATION:VALIDATE-CONFIGURATION-FORM in DEFUN
WARNING: redefining UIOP/CONFIGURATION:VALIDATE-CONFIGURATION-FILE in DEFUN
WARNING: redefining UIOP/CONFIGURATION:VALIDATE-CONFIGURATION-DIRECTORY in DEFUN
WARNING: redefining UIOP/CONFIGURATION:RESOLVE-RELATIVE-LOCATION in DEFUN
WARNING: redefining UIOP/CONFIGURATION:RESOLVE-ABSOLUTE-LOCATION in DEFUN
WARNING: redefining UIOP/CONFIGURATION:LOCATION-DESIGNATOR-P in DEFUN
WARNING: redefining UIOP/CONFIGURATION:LOCATION-FUNCTION-P in DEFUN
WARNING:
   redefining UIOP/CONFIGURATION:REGISTER-CLEAR-CONFIGURATION-HOOK in DEFUN
WARNING: redefining UIOP/CONFIGURATION:CLEAR-CONFIGURATION in DEFUN
WARNING: redefining UIOP/CONFIGURATION:UPGRADE-CONFIGURATION in DEFUN
WARNING: redefining UIOP/CONFIGURATION:GET-FOLDER-PATH in DEFUN
WARNING: redefining UIOP/CONFIGURATION:XDG-DATA-HOME in DEFUN
WARNING: redefining UIOP/CONFIGURATION:XDG-CONFIG-HOME in DEFUN
WARNING: redefining UIOP/CONFIGURATION:XDG-DATA-DIRS in DEFUN
WARNING: redefining UIOP/CONFIGURATION:XDG-CONFIG-DIRS in DEFUN
WARNING: redefining UIOP/CONFIGURATION:XDG-CACHE-HOME in DEFUN
WARNING: redefining UIOP/CONFIGURATION:XDG-RUNTIME-DIR in DEFUN
WARNING: redefining UIOP/CONFIGURATION:SYSTEM-CONFIG-PATHNAMES in DEFUN
WARNING: redefining UIOP/CONFIGURATION:FILTER-PATHNAME-SET in DEFUN
WARNING: redefining UIOP/CONFIGURATION:XDG-DATA-PATHNAMES in DEFUN
WARNING: redefining UIOP/CONFIGURATION:XDG-CONFIG-PATHNAMES in DEFUN
WARNING: redefining UIOP/CONFIGURATION:FIND-PREFERRED-FILE in DEFUN
WARNING: redefining UIOP/CONFIGURATION:XDG-DATA-PATHNAME in DEFUN
WARNING: redefining UIOP/CONFIGURATION:XDG-CONFIG-PATHNAME in DEFUN
WARNING: redefining UIOP/CONFIGURATION::COMPUTE-USER-CACHE in DEFUN
WARNING: redefining UIOP/BACKWARD-DRIVER:COERCE-PATHNAME in DEFUN
WARNING: redefining UIOP/PATHNAME:COMPONENT-NAME-TO-PATHNAME-COMPONENTS in DEFUN
WARNING: redefining UIOP/CONFIGURATION:USER-CONFIGURATION-DIRECTORIES in DEFUN
WARNING: redefining UIOP/CONFIGURATION:SYSTEM-CONFIGURATION-DIRECTORIES in DEFUN
WARNING: redefining UIOP/CONFIGURATION:IN-FIRST-DIRECTORY in DEFUN
WARNING: redefining UIOP/CONFIGURATION:IN-USER-CONFIGURATION-DIRECTORY in DEFUN
WARNING:
   redefining UIOP/CONFIGURATION:IN-SYSTEM-CONFIGURATION-DIRECTORY in DEFUN
```

attila-lendvai:
i think this is a bug in how UIOP is packaged into SBCL, or at least an unfortunate solution.

```
$ rlwrap sbcl --no-userinit
This is SBCL 1.2.4.debian, an implementation of ANSI Common Lisp.
More information about SBCL is available at <http://www.sbcl.org/>.

SBCL is free software, provided as is, with absolutely no warranty.
It is mostly in the public domain; some portions are provided under
BSD-style licenses.  See the CREDITS and COPYING files in the
distribution for more information.
* (require :asdf)

("ASDF" "asdf" "UIOP" "uiop")
* (asdf:component-loaded-p :uiop)

NIL
* (asdf:component-pathname (asdf:find-system :uiop))

#P"/home/alendvai/common-lisp/asdf/uiop/"
* 
```

which means that an `(asdf:load-system :uiop)` will reload/upgrade `uiop` from my own install of it, leading to the redefinition warnings.

let's ask @fare what he has to say about this!

fare:
Since every implementation now bundles ASDF 3.1, and ASDF transcludes UIOP, and depending on UIOP can only make things worse, I recommend never explicitly depending on UIOP. If a user is missing it, tell them to upgrade their implementation, or if they can't, to install a newer ASDF on top of it.

fare:
Also, recent changes I made to CFFI already depend on ASDF 3, so UIOP is guaranteed to be present.

luismbo:
Should/could ASDF mark uiop as loaded as Attila (and I) expected?

fare:
ASDF *does* mark UIOP as already loaded — but only if it can't find the system. If it can find the system, it assumes the version you installed and configured has precedence.

Now, the only one who installs and configures UIOP without ASDF is... Quicklisp!

Somehow,  Xach took an aversion to ASDF 3, and decided to keep shipping Quicklisp with ASDF 2.26 (last in the ASDF 2 series), though it's not pervert enough so as to downgrade when the implementation already provides ASDF 3 — which every single implementation now does (happily). But Quicklisp is still shipping UIOP for the sake of applications that depend on UIOP but not ASDF 3.

Maybe one of you can convince Xach to embrace ASDF 3, and either ship a more recent version (he hasn't upgraded since 2013) or just rely on what implementations provide. Then he doesn't need to ship UIOP as a separate package.

luismbo:
OK, thanks for your explanation! Should be fixed now.

jimka2001:
Sorry, I didn't understand what change was made in cffi to fix this?  Or was it closed, deciding to change nothing?  Sorry, I'm just confused.  Please make an explicit comment.

luismbo:
https://github.com/cffi/cffi/commit/47db2153670d9b46ee53a0deecb63e7248819cd8 should fix it.

fare:
The fix is to not explicitly depend on uiop, but implicitly depend on it.

If anyone uses an ancient implementation with ASDF 2.26 or older, they need to upgrade their implementation and/or ASDF, or they deserve to lose.

ASDF 3.1 has been out for > 2 years, ASDF 3.0 for > 3 years.

dkochmanski:
To add my 2 cents, UIOP is system dependency, so I agree with @attila-lendvai on that it should be listed there.

Also, not listing it breaks the `program-op` builds, if the user will decide to skip asdf/uiop in the final executable with `:nouiop t` option, or wants to builds only with uiop, i.e without `asdf`.  Can we reopen this, or it's already set in stone?

dkochmanski:
Moreover (and this isn't implemented in ASDF yet afaik), if you don't want to reload UIOP to the new version, then it should be listed in dependencies as `(:require #:uiop)` – right now it just makes it an ordinary dependency, but it will be future compatible with `request-system` as defined in the manual:

https://common-lisp.net/project/asdf/asdf.html#index-require_002dsystem

luismbo:
@dkochmanski thanks for your input. I'm not sure what the fix should be now. Just revert the last commit so that uiop is listed again in the dependencies or should we do something else?

dkochmanski:
I would list it as `(:require :uiop)` – I'll send a PR in a second. For now it will just load `uiop` as normal dependency, but it may be fixed in `ASDF` in near future.

fare:
`:nouiop t` is an ECL-specific escape to `program-op`. If you're using it, you're supposed to know what you're doing, at which point it's your fault if your program is using CFFI and fails to link because CFFI depends on UIOP. Many other systems implicitly depend on UIOP, or sometimes on ASDF itself!

That's not a CFFI specific issue, and it's a bad idea of CFFI to make the experience less for every user of every platform just for the sake of a few advanced ECL users who should know better than to specify `:nouiop t` and expect things to just work without testing.

fare:
And no, `(require :uiop)` is a no-op when building with ASDF.

dkochmanski:
I don't mean calling `(require :uiop)`, but declaring it that way:

```common-lisp
(defsystem #:xxx
  :depends-on ((:require :uiop)))
```

dkochmanski:
I think it's a bad idea to make implicit dependencies and if any system depends on uiop implicitly it's rather a bug than a feature

fare:
As long as Quicklisp ships an antique ASDF and then ships UIOP independently from ASDF, that's a feature.

ECL is the only platform for which that matters, and only if you create an image with `:nouiop t`. I don't see why everyone should bend over backwards to support this platform. Moreover, your solution doesn't even help, since it doesn't affect at all the code that ECL will link (which can be considered a separate bug in ASDF, but oh well).

dkochmanski:
I can hardly imagine how it's a feature. For now we have a situation as follows:

Common Lisp implementation which ships antique ASDF *can* load UIOP and CFFI (which supports for instance clisp, corman and gcl), while with this indirect dependency it breaks these systems.

From the system dependency point of view – ASDF is meant as a tool to define and build systems, and because it needs various features, it brings UIOP on board – that's fine. If anyone wants to bundle the whole build-system with the final executable, then it's OK, but usually it's not necessary. Fact that most CL dump images of the whole process doesn't automatically render it obvious. Application shouldn't depend on its whole build system at runtime.

Declaring explicitly dependencies is important for maintainable systems and it's no more bending backwards than introducing messy practices, where the system definition doesn't actually define the system, but parts of it, that in consequence breaks the promise, that you may build system based on its description.

Moreover there is no reason (except bug in ASDF, on which I'm working right now) to not list a dependency as a dependency.  Fixing bug in ASDF by making people introduce bug into their system definition isn't a way to go.

fare:
If you're trying to use a mix of ancient implementations and modern libraries, then explicitly provide a non-standard option that doesn't work in the combination, it's your problem.

Why do you expect `:nouiop t` to either work or be useful in this case??? If you don't use it, there is no problem.

fare:
Each and every piece of software that uses ASDF to build ipso facto depends on ASDF and UIOP. There isn't any sense in making the dependency explicit unless there is an alternative build system, e.g. Bazel.

dkochmanski:
`no-uiop t` option was an example of how to build a final executable without `asdf` or `uiop` in it – executable is smaller for instance, or the programmer doesn't want software he won't use (or doesn't like). The problem though is more general and I have already enumerated by concerns, ranging from clean system definition, through maintainable software to the simple fact, that not listing dependencies (making them implicit) is plain wrong.

There may be Bazel or some other build system in the future. Do you expect, that this potential build system will ship UIOP, whenever systems need it or not? While UIOP makes a perfect case for a library, it isn't in a shape to mandate new standard, so systems with implicit UIOP dependency will be broken in the future (unlike many old CL systems created before ASDF, which work just fine with today implementations).

Moreover, there is absolutely no valid reason to not list it in dependencies (bugs in build system doesn't count as a valid reason).

dkochmanski:
`Each and every piece of software that uses ASDF to build ipso facto depends on ASDF and UIOP.` – at build time. Like every piece of software that uses Makefile to build ipso facto depens on Makefile and autotools. That doesn't mean that people actually ship Makefile/autotools with their application in one executable. This isn't a perfect analogy because CL is so much more and you can actually include it, but you got a drill – it's not that you always want to.

fare:
Bazel builds are already possible, and are not described by `.asd` files but by BUILD files. There's a CFFI definition in the bazelisp repository, and it indeed includes an explicit dependency on UIOP. Bazelisp doesn't currently support ECL, but I suppose you could make that happen if you wanted.

ASDF is not at all geared for cross-compilation. ECL using `:no-uiop t` is the only, limited, instance, that explicitly defeats the ASDF implicit linking of UIOP, at which point you're on your own.

fare:
If you want cross-compilation support in ASDF, it's possible, but will be lots of non-trivil changes, and pain to debug on 15+ implementations. See the ASDF TODO for some details.

In the meantime, it's absurd to make everyone suffer for the sake of a feature that doesn't exist.

dkochmanski:
I've already enumerated my concerns, and I believe they are all valid and ECL is only small aspect of the problem, and I believe you've noticed that. I won't repeat them, since we have a backlog.

P.S. Regarding CC, I already have semi-working module for that as ASDF extension, but it's not quite ready yet.

P.S.S. I know worse ways to suffer than listing dependencies of my system :-)

fare:
I propose you first fix ASDF or get it fixed, and then there will be something to be done about CFFI.

PR #99, opened 2016-09-09
system: re-add dependency on UIOP as :require

dkochmanski:
Dependency on `uiop' can't be implicit for sake of `program-op' and
`image-op' builds. User may wish to not include `asdf' or `cmp' (in case
of ECL) in the final image.

ASDF has option (require-system :system) which doesn't reload it, if
it's already present in the image (even if it's newer):

https://common-lisp.net/project/asdf/asdf.html#index-require_002dsystem

so (:require :uiop) shouldn't reload the system. That said, ASDF doesn't
handler (:require :system-name) options correclty yet, treating them
like normal dependencies. But this will be forward-compatible when ASDF
will fix that.

Closes #96.

### DIFF
--- a/cffi.asd
+++ b/cffi.asd
@@ -36,7 +36,7 @@
   :author "James Bielman  <jamesjb@jamesjb.com>"
   :maintainer "Luis Oliveira  <loliveira@common-lisp.net>"
   :licence "MIT"
-  :depends-on (:alexandria :trivial-features :babel)
+  :depends-on (:uiop :alexandria :trivial-features :babel)
   :in-order-to ((test-op (load-op :cffi-tests)))
   :perform (test-op (o c) (operate 'asdf:test-op :cffi-tests))
   :components


fare:
I can't imagine how this actually helps in any way.

dkochmanski:
see the comment below the issue #96

luismbo:
Having skimmed the documentation you pointed to, it seems like
`require-system` is geared towards interactive usage. Now, that you've
convinced me that UIOP is just like any other library (that therefore
should be listed along with other dependencies), I don't understand what's
different about it that makes us want to forego loading the most recent
available version, as is done for any other library.

Seems like it might be confusing if I were developing CFFI against some
recent version of UIOP (particularly if patched by myself).

dkochmanski:
Right, I would go with a library dependency in that case. `(:require :uiop)` dependency would preserve the behaviour of making `:uiop` dependency implicit (when ASDF is fixed), while leaving it as `:uiop` will load the most recent version. In that case you should revert the commit which removed the direct dependency.

Regarding the redefinition warnings, I'll think how to muffle them if the system is preloaded (I'm fixing `require-system` mechanism in ASDF right now, so I can look into that as well).

attila-lendvai:
re "loading the most recent available version": does UIOP also do the version comparison and the smart self-upgrade that ASDF does? if yes then what we are talking about is merely a few redefinition warnings (in case there's a newer version available), which are not nice, but not the end of the world either.

dkochmanski:
problem is a bit deeper than that, because user may modify uiop locally without bumping a version and user rightfully expects, that when he loads the new uiop (i.e linked in `local-projects` directory), then it will replace the previous version. Also, if he modify it (because he is hacking, not a release yet), then it should reload, otherwise it shouldn't take any action.

To do that, ASDF compares timestamps of the files, but the preloaded uiop module doesn't have a timestamp, so loading UIOP (unless it will be `request-system` or uiop is marked as `immutable`), will replace it in the memory.

You are either depending on the bundled module (then you should do `(:require :uiop)` in the dependency list – this doesn't work yet though and you load the library anyway) or you depend on the library accessible through the registry.

But I agree, that these redefinition warnings aren't dangerous, because if it's listed in the dependencies, then developer also tested the library with the `uiop` library, not a bundled system (so even if quicklisp ships older `uiop`, the older version is what the library was tested with anyway).

fare:
UIOP is special because ASDF 3 itself depends on it, transcludes it, and thus anything that builds with ASDF depends on it and has it already. (Note: ASDF 2 officially died last year when LispWorks eventually upgraded to ASDF 3.1).

Quicklisp distributed UIOP separately because it did not want to adopt ASDF 3. Not useful anymore, but still an inconvenience.

fare:
@attila-lendvai no, uiop does not do the magic anti-downgrade check like asdf does.

Indeed, it's hard enough to have one magic kludge for asdf, and the uiop API is hopefully stable and monotonic enough, and separated enough from the guts of the asdf data structures, that it shouldn't be as critical not to downgrade a bit (especially if the old functions remain defined for the sake of code previously using it).

That doesn't mean no harm can come from downgrading uiop, but the harm will hopefully be rarer, lesser, and easier to pinpoint on those who did it.

fare:
Speaking of downgrade of UIOP, I just realized Quicklisp had failed to pick up uiop 3.1.7, because Robert's new old-style Makefile used .tgz instead of .tar.gz for released files. Sigh.

dkochmanski:
@luismbo adjusted the PR to add `uiop` as a simple dependency.

luismbo:
Applied. Thanks for the great commit message.

PR #100, opened 2016-09-22
c2ffi now exports ns tags for function & constant

cbaggers:
I was getting warnings about `:ns` from function and constant so I have added them to the 'expected' lists.

### DIFF
--- a/src/c2ffi/generator.lisp
+++ b/src/c2ffi/generator.lisp
@@ -655,7 +655,7 @@ target package."
   #+nil
   (assume (equal "extern" storage-class)
           "Unexpected function STORAGE-CLASS: ~S for function ~S" storage-class function-name)
-  (expected-json-keys -json-entry- :tag :name :return-type :parameters :variadic :inline :storage-class)
+  (expected-json-keys -json-entry- :tag :name :return-type :parameters :variadic :inline :storage-class :ns)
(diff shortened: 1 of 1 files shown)


attila-lendvai:
what's in that tag, the C++ namespace?

could you post a few examples?

do you think it's safe to just ignore it, or shall we put an assert for the base C namespace that fails with any C++ namespace magic? (if i'm not completely lost here)

cbaggers:
Good questions. I shouldn’t have really opened the PR for this yet. I will get answers to those questions and get back to you here.

cbaggers:
Ok so it does appear to be for the namespace. Although, in the project where I found the issue, the value was always `0`. I'm guessing this was `NULL` in c2ffi.

It also seems we already have support for `ns` in a couple of places in `cffi/c2ffi`:
- `%process-struct-like`
- `typedef`

they were added in the big https://github.com/cffi/cffi/pull/84/commits/2a792286129baca116bee4ad8b03aa3a5481ca03 commit

So far neither of those two seem to do anything with that data.

cbaggers:
I have also had a read into `cl-autowrap` and it seems it does not use the `ns` info yet either. Because of this I think we can safely ignore it

attila-lendvai:
merged, thank you!

PR #102, opened 2016-09-24
Add caching of grovel & wrap results

cbaggers:
So the goal is to be able to load the wrapper & .so files from a cache
rather than having to build them on every person's platform.

The interesting thing about the groveller is that we are allowed to put
reader conditionals (#+ & #-) in the wrapper specification. This means
that a cached result is only relevent if all the results of all the
conditionals match. To find out the required features (feature-set) we
wrap the specification loading in a with-cached-reader-conditionals
form.

So now if you specify a :cache-dir for a :soname wrapper the system
will now check for the existance of a feature specific folder within
cache-dir. If that dir exists it will copy .so and wrappers files to
the destination folder rather than making the c lib.

The name of the 'feature specific folder' we talked about above is
generated from the feature-set we talked about above (see note on cache
name below)

When loading from cache we use #'touch-file on the c-file and
o-file that would have been generated. This is to stop the warning that
results from annoucing a file in #'output-files and not creating it.

## Cache Name

As mentioned above we need a feature specfic name for the cache. To do
this we take the following:

- (or (uiop:operating-system) (software-type))
- (or (uiop:architecture) (machine-type))
- the features referenced by the reader conditionals in the spec file
  these are concatenated and hashed

And the we concatenate them into a string as follows

    (format nil "~a_~a_~x" os architecture hashed-features)

This works well except for the fact that it doesnt capture the
distinctions between different versions of the OS (e.g. win7 v win8). It
would be great to add a function to get that and then replace `os` with
that string.

## Side Note

So before we had code which generated the names of the output files
in the #'output-files method and then generated the names again for
the code doing the work.

This was just asking for bugs and was a waste of work so we now
destructure the names from #'output-files in the #'perform methods and
pass these names to the various worker functions

However this new version of `#'process-grovel-file` has a different
signature which could cause problems with backwards compatibility. To
that end I have named the new version `process-grovel-file*` and
redefined the old `process-grovel-file` to call the new one.

### DIFF
--- a/cffi-grovel.asd
+++ b/cffi-grovel.asd
@@ -32,10 +32,12 @@
   :licence "MIT"
   :components
   ((:module "grovel"
+    :serial t
     :components
     ((:static-file "common.h")
      (:file "package")
-     (:file "grovel" :depends-on ("package"))
-     (:file "asdf" :depends-on ("grovel"))))))
+     (:file "feature-caching-read")
+     (:file "grovel")
+     (:file "asdf")))))
 
 ;; vim: ft=lisp et
--- a/grovel/asdf.lisp
+++ b/grovel/asdf.lisp
@@ -75,7 +75,7 @@
 ;;;# ASDF component: GROVEL-FILE
 
 (defclass grovel-file (process-op-input cc-flags-mixin)
-  ()
+  ((cache-dir :initform nil :initarg :cache-dir :accessor cache-dir-of))
   (:default-initargs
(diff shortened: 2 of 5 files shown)


cbaggers:
Ouch, well that failed the tests. I'll get on those now

cbaggers:
Hmm reading this again I'm not happy the the 2 places I just swallow potential errors :| I’m going to remove that.

PR #103, opened 2016-09-28
Add ffi-form-generators feature to c2ffi

cbaggers:
The goal of this feature is to enable people to emit their own code
based on the generated form.

To use this you specify one or more generator forms using the
`:ffi-form-generators` key in your `:cffi/c2ffi-file` component.

For example:

```
:ffi-form-generators ((:defcstruct "raw-bindings-nuklear.ffi::ffi-generator-struct"))
```

This means that for every `cffi:defcstruct` form that is written to the
generated file `raw-bindings-nuklear.ffi::ffi-generator-struct` will be
called. Whatever form is returned from that function will be written
into the generated file using `#'output/code`.

The first element of the form (:defcstruct in the above case) is called
the 'pattern' and the second is the 'generator function'.

You can define multiple generator forms:

```
:ffi-form-generators ((:defcstruct "raw-bindings-nuklear.ffi::ffi-generator-struct")
                      (:defcunion "raw-bindings-nuklear.ffi::ffi-generator-union"))
```

It is legal to have both call the same function if you like.

These are the valid patterns you can use in your generator forms:

```
:defcfun
:defcstruct
:defcunion
:defctype
:defcvar
:defbitfield
:defcenum
:begin
:end
```

A generator function using the `:begin` pattern will be called once,
just before the c2ffi entries are processed

A generator function using the `:begin` pattern will be called once,
just after the c2ffi entries are processed

The signature for your generator function should be as follows:

```
(defun your-function (definition context &key &allow-other-keys)
  ...)
```

The context argument will be an instance of the `ffi-generator-context`
class. This object has one field called data, it has an accessor called
`#'context-data`.

The context exists for cases where you want to aggregate data during the
processing, usually with the goal of emitting it at the `:end`.

This can be useful when you want to make sure you don't emit forms for
duplicate definitions of cffi types. For example you may want to emit an
accessor function for cetain CFFI structs, however SBCL will throw an
error if it finds duplicate `defun`s in the same file

Here is an example of how this feature can be used. The goal is to emit
a `defstruct` form for every `cffi:defcstruct`.

In the asd file:

```
:ffi-form-generators ((:begin "raw-bindings-nuklear.ffi::ffi-generator-begin")
                  (:defcstruct "raw-bindings-nuklear.ffi::ffi-generator-struct")
                  (:end "raw-bindings-nuklear.ffi::ffi-generator-end"))
```

In a auxilary lisp file:

```
;; we make a hash table to store and deduplicate the defstruct forms
(defun ffi-generator-begin (definition context &key &allow-other-keys)
  (declare (ignore definition))
  (setf (context-data context) (make-hash-table :test #'equal))
  nil)

;; For each defcstruct we make a defstruct form and store it in the
;; context
(defun ffi-generator-struct (definition context &key &allow-other-keys)
  (destructuring-bind (name-and-options . fields) (rest definition)
    (destructuring-bind (name . options) (uiop:ensure-list name-and-options)
      (declare (ignore options))
      (when fields
        (let* ((wrapper-name-str (format nil "~a-WRAPPER" name))
               (wrapper-name (intern wrapper-name-str)))
          (setf (gethash `(:struct ,name) (context-data context))
                `(defstruct ,wrapper-name ,@(mapcar #'first fields)))))))
  nil)

;; We emit all the defstruct forms
(defun ffi-generator-end (definition context &key &allow-other-keys)
  (declare (ignore definition))
  (let ((data (context-data context)))
    `(progn
       ,@(loop :for k :being :the :hash-keys :of data :collect
            (gethash k data)))))
```

This feature is slightly more open ended than the other :ffi-*
transformer and export forms, however I feels it's flexibility worth the
small cost in difficulty.

### DIFF
--- a/src/c2ffi/asdf.lisp
+++ b/src/c2ffi/asdf.lisp
@@ -54,8 +54,8 @@
                          :initform 'default-ffi-type-transformer)
    (ffi-export-predicate :initarg :ffi-export-predicate
                          :initform 'default-ffi-export-predicate)
-   (ffi-form-generators :initarg :ffi-form-generators
-                        :initform nil)
+   (ffi-form-generator :initarg :ffi-form-generator
+                        :initform 'default-ffi-form-generator)
    (foreign-library-name :initarg :foreign-library-name
                          :initform nil)
    (foreign-library-spec :initarg :foreign-library-spec
@@ -178,7 +178,7 @@ file, except that it's will be stored in the fasl cache."))
                :for arg :in '(ffi-name-transformer
                               ffi-type-transformer
                               ffi-export-predicate
-                              ffi-form-generators
+                              ffi-form-generator
                               foreign-library-name
                               foreign-library-spec
                               emit-generated-name-mappings
--- a/src/c2ffi/asdf.lisp
+++ b/src/c2ffi/asdf.lisp
@@ -54,6 +54,8 @@
                          :initform 'default-ffi-type-transformer)
    (ffi-export-predicate :initarg :ffi-export-predicate
                          :initform 'default-ffi-export-predicate)
+   (ffi-form-generators :initarg :ffi-form-generators
+                        :initform nil)
    (foreign-library-name :initarg :foreign-library-name
                          :initform nil)
    (foreign-library-spec :initarg :foreign-library-spec
@@ -176,6 +178,7 @@ file, except that it's will be stored in the fasl cache."))
                :for arg :in '(ffi-name-transformer
                               ffi-type-transformer
                               ffi-export-predicate
+                              ffi-form-generators
                               foreign-library-name
                               foreign-library-spec
                               emit-generated-name-mappings
--- a/src/c2ffi/asdf.lisp
+++ b/src/c2ffi/asdf.lisp
@@ -54,8 +54,8 @@
                          :initform 'default-ffi-type-transformer)
    (ffi-export-predicate :initarg :ffi-export-predicate
                          :initform 'default-ffi-export-predicate)
-   (ffi-form-generator :initarg :ffi-form-generator
-                        :initform 'default-ffi-form-generator)
+   (ffi-plugin-factory :initarg :ffi-plugin-factory
+                        :initform 'default-ffi-plugin-factory)
    (foreign-library-name :initarg :foreign-library-name
                          :initform nil)
    (foreign-library-spec :initarg :foreign-library-spec
@@ -178,7 +178,7 @@ file, except that it's will be stored in the fasl cache."))
                :for arg :in '(ffi-name-transformer
                               ffi-type-transformer
                               ffi-export-predicate
-                              ffi-form-generator
+                              ffi-plugin-factory
                               foreign-library-name
                               foreign-library-spec
                               emit-generated-name-mappings
--- a/src/c2ffi/generator.lisp
+++ b/src/c2ffi/generator.lisp
@@ -57,6 +57,7 @@
 ;; Called on the CFFI type, e.g. to turn (:pointer :char) into a :string.
 (defvar *ffi-type-transformer* 'default-ffi-type-transformer)
 (defvar *ffi-export-predicate* 'default-ffi-export-predicate)
+(defvar *ffi-form-generators* nil)
 
 (define-constant +generated-file-header+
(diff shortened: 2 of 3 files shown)


cbaggers:
@attila-lendvai : Hi mate, I expect this will be one you'll be wanting to review :)

attila-lendvai:
hrm. did you consider using CLOS multiple dispatch? if you did why not that?

the reason i'm asking is that one may want to generate some aux defun for each :defcstruct (BTW, i don't like the name, i'd stick to the C names here), and then (call-next-method) can be useful.

also, i'm not convinced that the added complexity is justified for this feature. can you please show some real-world examples of what you use it for?

if you just want to generate extra definitions then maybe it would be better to give a registry of all the emitted definitions to one user hook which can then generate whatever it wants?

cbaggers:
Sorry for the confusion, my description could have been clearer. I didn't pick clos for two reasons, it didn't match the style of the code already in the file, and (to me) didn't make the behaviour any clearer.

I'm a little confused by the " one may want to generate some aux defun for each :defcstruct" as that works already. The context is only relevant if you want to collect things for the :end.

The :defcstruct name was used as the user is matching against the form being emitted into the file and that's the cffi name.

However I'm happy to rework this feature if you like. I'd like to be able to do what I can do now though (dedup and emit at the end *and* emit for every matching form.)

I'm afraid I'm writing this from my mobile but I can give examples for each use case when I'm home.

Thanks for the input

cbaggers:
@attila-lendvai thank again for your review yesterday. I looked at my work again today and agreed that it was too complicated. I have rewritten it to be much simpler. Here is the new description:

## ffi-form-generator feature

The ffi-form-generator argument allows you to specify a function to be
called once for every form that is written into the generated lisp file.

The generator function is specified like this:

    :ffi-form-generator "function-name"

any non nil value returned from the generator function will be written
into the generated lisp file with #'output/code.

An example function would be:

    (defun ffi-form-generator (definition &key &allow-other-keys)
      (let ((def-name (second definition)))
        (when (eq def-name 'defcstruct)
          (let* ((name (intern (format nil "~a-WRAPPER" def-name))))
            `(defstruct ,name (mapcar #'first (cddr definition)))))))

This will emit one lisp wrapper struct for every C Struct emitted.

You can optionally also return a second value that will be used as a
deduplication-key. CFFI will collect all of the forms returned with a
deduplication-key, remove the duplicates, and emit all the forms at the
end of the generated lisp file.

For example if we decided that we want to make sure that our earlier
example does not emit multiple defstructs if there are multiple C
structs we can do the following.

    (defun ffi-form-generator (definition &key &allow-other-keys)
      (let ((def-name (second definition)))
        (when (eq def-name 'defcstruct)
          (let* ((name (intern (format nil "~a-WRAPPER" def-name))))
            (values `(defstruct ,name (mapcar #'first (cddr definition)))
                    def-name)))))

attila-lendvai:
it's not forgotten, just waiting for some hacking time.

meanwhile, what do you think about an alternative API where there's, let's say a `plugin-factory` or somesuch that is supposed to return two closures:

1. one that will be called with each generated form
2. and one that will be called at the end of the generation?

this way they could both capture a hashtable and keep even more complexity out of the codebase.

i may hack it up sometime, or if you feel inspired feel free to take a stab at it, but let's keep each other informed here, so that neither of us work in vain.

cbaggers:
Thanks for the update :) Yeah that works well too and keeps the user very much in control.
I'll definitely have a go at this when I next get some good hacking time.

> but let's keep each other informed here, so that neither of us work in vain
Agreed.

cbaggers:
Turns out I have some time tonight. I'll push a first version soon.

cbaggers:
Ok, here is a first pass at this.

attila-lendvai:
i have pushed some cleanups of this into https://github.com/attila-lendvai/cffi/commits/form-generators

i squashed your three commits, and then recorded my cleanup as a separate one. let me know what you think, and if it's fine i'll push it to master.

cbaggers:
@attila-lendvai Looks good to me :)

attila-lendvai:
ok, pushed to master. thanks!

luismbo:
Thanks @attila-lendvai and @cbaggers.

PR #105, opened 2016-10-03
better error for in-package grovelling

cbaggers:
Before the error was along the lines of `nil is not a package` which made it a little harder to track down which declaration was incorrect

### DIFF
--- a/grovel/grovel.lisp
+++ b/grovel/grovel.lisp
@@ -802,8 +802,12 @@ string."
     (process-wrapper-form out form)))
 
 (define-wrapper-syntax in-package (name)
-  (setq *package* (find-package name))
-  (push `(in-package ,name) *lisp-forms*))
+  (let ((desired-package (find-package name)))
+    (assert desired-package ()
+            "Wrapper file specified (in-package ~s)~%however that does not name a known package"
+            name)
+    (setq *package* desired-package)
+    (push `(in-package ,name) *lisp-forms*)))
 
 (define-wrapper-syntax c (&rest strings)
   (dolist (string strings)
--- a/grovel/grovel.lisp
+++ b/grovel/grovel.lisp
@@ -802,10 +802,10 @@ string."
     (process-wrapper-form out form)))
 
 (define-wrapper-syntax in-package (name)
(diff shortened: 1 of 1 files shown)


grovel/grovel.lisp
```diff
@@ -802,8 +802,12 @@ string."
     (process-wrapper-form out form)))
 
 (define-wrapper-syntax in-package (name)
-  (setq *package* (find-package name))
-  (push `(in-package ,name) *lisp-forms*))
+  (let ((desired-package (find-package name)))
+    (assert desired-package ()
```
luismbo: Do you think it'd be worthwhile to have `(assert (find-package name) (name) ...` instead? That'd provide a restart for trying another package name. Probably overkill...

grovel/grovel.lisp
```diff
@@ -802,8 +802,12 @@ string."
...
+            "Wrapper file specified (in-package ~s)~%however that does not name a known package"
```
luismbo: Do you think the sentence `"The name ~s does not designate any package."` would be too terse? (Borrowed from SBCL.)

attila-lendvai: more diverse error messages are easier to search for. probably not a big deal as the users are programmers and should get a backtrace, but still...

attila-lendvai: i've been badly in need of a functional backtrace often enough to think of that^ response...

cbaggers: @luismbo I like it having 'Wrapper' at the beginning as it sets the context for the rest of the message quickly. I generally want to mentally narrow down an issue as fast as possible

grovel/grovel.lisp
luismbo: Yes, assert evaluates its first argument again to check if it has become true.

grovel/grovel.lisp
luismbo: Good points. Then, just end then sentence with a period and I think it's perfect.

cbaggers:
I have added the suggestions

luismbo:
Applied as d40df4d with minor refactoring. Thanks!

PR #106, opened 2016-10-03
Add caching of grovel & wrap results

cbaggers:
So the goal is to be able to load the wrapper & .so files from a cache
rather than having to build them on every person's platform.

The interesting thing about the groveller is that we are allowed to put
reader conditionals (#+ & #-) in the wrapper specification. This means
that a cached result is only relevent if all the results of all the
conditionals match. To find out the required features (feature-set) we
wrap the specification loading in a with-cached-reader-conditionals
form.

So now if you specify a :cache-dir for a :soname wrapper the system
will now check for the existance of a feature specific folder within
cache-dir. If that dir exists it will copy .so and wrappers files to
the destination folder rather than making the c lib.

The name of the 'feature specific folder' we talked about above is
generated from the feature-set we talked about above (see note on cache
name below)

When loading from cache we use #'touch-file on the c-file and
o-file that would have been generated. This is to stop the warning that
results from annoucing a file in #'output-files and not creating it.

Cache Name
----------

As mentioned above we need a feature specfic name for the cache. To do
this we take the following:

- (or (uiop:operating-system) (software-type))
- (or (uiop:architecture) (machine-type))
- the features referenced by the reader conditionals in the spec file
  these are concatenated and hashed

And the we concatenate them into a string as follows

    (format nil "~a_~a_~x" os architecture hashed-features)

This works well except for the fact that it doesnt capture the
distinctions between different versions of the OS (e.g. win7 v win8). It
would be great to add a function to get that and then replace `os` with
that string.

Side Notes
----------

`[0]` So before we had code which generated the names of the output files
in the #'output-files method and then generated the names again for
the code doing the work.

This was just asking for bugs and was a waste of work so we now
destructure the names from #'output-files in the #'perform methods and
pass these names to the various worker functions

However this new version of `#'process-grovel-file` has a different
signature which could cause problems with backwards compatibility. To
that end I have named the new version `process-grovel-file*` and
redefined the old `process-grovel-file` to call the new one.

`[1]` The wrapper also used to always use absolute paths for the search-path
in the generated 'processed-wrapper-file'. This would cause problem if
you were to shipped the cached files. So now, if the files are being
cached, we write the path as a call to asdf:system-relative-path

`[2]` A variant of `include` called `include-local` has been added to
make wrapping of single-header-file libraries easier. The arguments are
asdf paths.

### DIFF
--- a/cffi-grovel.asd
+++ b/cffi-grovel.asd
@@ -32,10 +32,12 @@
   :licence "MIT"
   :components
   ((:module "grovel"
+    :serial t
     :components
     ((:static-file "common.h")
      (:file "package")
-     (:file "grovel" :depends-on ("package"))
-     (:file "asdf" :depends-on ("grovel"))))))
+     (:file "feature-caching-read")
+     (:file "grovel")
+     (:file "asdf")))))
 
 ;; vim: ft=lisp et
--- a/grovel/asdf.lisp
+++ b/grovel/asdf.lisp
@@ -75,7 +75,7 @@
 ;;;# ASDF component: GROVEL-FILE
 
 (defclass grovel-file (process-op-input cc-flags-mixin)
-  ()
+  ((cache-dir :initform nil :initarg :cache-dir :accessor cache-dir-of))
   (:default-initargs
    :generated-lisp-file-type "processed-grovel-file")
   (:documentation
@@ -86,22 +86,25 @@
   (let* ((input-file (first (input-files op c)))
          (output-file (make-pathname :type (generated-lisp-file-type c)
                                      :defaults input-file))
-         (c-file (make-c-file-name output-file "__grovel")))
+         (c-file (make-c-file-name output-file "__grovel"))
+         (exe-file (make-exe-file-name c-file)))
     (list output-file
           c-file
-          (make-exe-file-name c-file))))
+          exe-file)))
+
 
 (defmethod perform ((op process-op) (c grovel-file))
-  (let* ((output-file (first (output-files op c)))
-         (input-file (first (input-files op c)))
-         (tmp-file (process-grovel-file input-file output-file)))
-    (rename-file-overwriting-target tmp-file output-file)))
+  (destructuring-bind (output-file c-file exe-file) (output-files op c)
+    (let* ((input-file (first (input-files op c)))
+           (absolute-cache-dir (absolute-cache-dir input-file (cache-dir-of c))))
+      (process-grovel-file* input-file output-file c-file exe-file absolute-cache-dir))))
 
 
 ;;;# ASDF component: WRAPPER-FILE
 
 (defclass wrapper-file (process-op-input cc-flags-mixin)
-  ((soname :initform nil :initarg :soname :accessor soname-of))
+  ((soname :initform nil :initarg :soname :accessor soname-of)
+   (cache-dir :initform nil :initarg :cache-dir :accessor cache-dir-of))
   (:default-initargs
    :generated-lisp-file-type "processed-wrapper-file")
   (:documentation
@@ -120,9 +123,10 @@
                                      :defaults input-file))
          (c-file (make-c-file-name output-file "__wrapper"))
          (o-file (make-o-file-name output-file "__wrapper"))
-         (lib-soname (wrapper-soname c)))
+         (lib-soname (wrapper-soname c))
+         (lib-file (make-lib-file-name (make-soname lib-soname output-file))))
     (list output-file
-          (make-lib-file-name (make-soname lib-soname output-file))
+          lib-file
           c-file
           o-file)))
 
@@ -135,19 +139,37 @@
         (multiple-value-bind (files translatedp) (call-next-method)
           (values (append files (list lib-file o-file)) translatedp)))))
 
+(defun absolute-cache-dir (input-file cache-dir)
+  (when cache-dir
+    (let ((input-dir (uiop:pathname-directory-pathname input-file)))
+      (uiop:ensure-directory-pathname
+       (uiop:subpathname input-dir cache-dir)))))
+
+(defun system-to-component-path (component)
+  (let* ((sys (component-system component))
+         (sys-dir (pathname-directory (component-pathname sys)))
+         (comp-dir (pathname-directory (component-pathname component))))
+    (make-pathname* :directory (cons :relative (subseq comp-dir (length sys-dir))))))
+
+(defun sys-relative-cache-dir (c)
+  (let ((cache-dir (cache-dir-of c)))
+    (when cache-dir
+      (ensure-directory-pathname
+       (subpathname* (system-to-component-path c) cache-dir)))))
+
 (defmethod perform ((op process-op) (c wrapper-file))
-  (let* ((output-file (first (output-files op c)))
-         (input-file (first (input-files op c)))
-         (tmp-file (process-wrapper-file
-                    input-file
-                    :output-defaults output-file
-                    :lib-soname (wrapper-soname c))))
-      (unwind-protect
-           (alexandria:copy-file tmp-file output-file :if-to-exists :supersede)
-        (delete-file tmp-file))))
+  (destructuring-bind (output-file lib-name c-file o-file) (output-files op c)
+    (let* ((input-file (first (input-files op c))))
+      (process-wrapper-file (component-system c)
+                            input-file
+                            output-file
+                            lib-name
+                            c-file
+                            o-file
+                            :lib-soname (wrapper-soname c)
+                            :cache-dir (sys-relative-cache-dir c)))))
 
 
 ;; Allow for naked :grovel-file and :wrapper-file in asdf definitions.
 (setf (find-class 'asdf::cffi-grovel-file) (find-class 'grovel-file))
 (setf (find-class 'asdf::cffi-wrapper-file) (find-class 'wrapper-file))
-
--- a/grovel/feature-caching-read.lisp
+++ b/grovel/feature-caching-read.lisp
@@ -1,0 +1,78 @@
+;;;; -*- Mode: lisp; indent-tabs-mode: nil -*-
+;;;
+;;; grovel.lisp --- The CFFI Groveller.
+;;;
+;;; Copyright (C) 2005-2006, Dan Knap <dankna@accela.net>
+;;; Copyright (C) 2005-2006, Emily Backes <lucca@accela.net>
+;;; Copyright (C) 2007, Stelian Ionescu <sionescu@cddr.org>
+;;; Copyright (C) 2007, Luis Oliveira <loliveira@common-lisp.net>
+;;;
+;;; Permission is hereby granted, free of charge, to any person
+;;; obtaining a copy of this software and associated documentation
+;;; files (the "Software"), to deal in the Software without
+;;; restriction, including without limitation the rights to use, copy,
+;;; modify, merge, publish, distribute, sublicense, and/or sell copies
+;;; of the Software, and to permit persons to whom the Software is
+;;; furnished to do so, subject to the following conditions:
+;;;
+;;; The above copyright notice and this permission notice shall be
+;;; included in all copies or substantial portions of the Software.
+;;;
+;;; THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+;;; EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+;;; MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+;;; NONINFRINGEMENT.  IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+;;; HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+;;; WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+;;; OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+;;; DEALINGS IN THE SOFTWARE.
+;;;
+
+(in-package #:cffi-grovel)
+
+(defun dirty-featurep (x)
+  "Ugly in implementation but will always match the implementations logic"
+  (with-standard-io-syntax
+    (with-input-from-string (s (format nil "#+~s t" x))
+      (read s nil nil))))
+
+(defun make-caching-reader-conditional ()
+  (let ((cache (make-array 0 :adjustable t :fill-pointer 0)))
+    (list (lambda (stream sub-char numarg)
+	    (declare (ignore numarg))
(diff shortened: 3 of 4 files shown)


cbaggers:
Ah I know this issue, it's me not supporting the old grovel signature for the tests. I'll fix that

cbaggers:
@luismbo done. This should be ready for pulling apart now :)

cffi-grovel.asd
```diff
@@ -32,10 +32,12 @@
   :licence "MIT"
   :components
   ((:module "grovel"
+    :serial t
```
cbaggers: save behaviour as before but a little tidier

grovel/asdf.lisp
```diff
@@ -86,22 +86,25 @@
   (let* ((input-file (first (input-files op c)))
          (output-file (make-pathname :type (generated-lisp-file-type c)
                                      :defaults input-file))
-         (c-file (make-c-file-name output-file "__grovel")))
+         (c-file (make-c-file-name output-file "__grovel"))
```
cbaggers: just a little tidying here

grovel/asdf.lisp
```diff
@@ -86,22 +86,25 @@
...
 
 (defmethod perform ((op process-op) (c grovel-file))
-  (let* ((output-file (first (output-files op c)))
-         (input-file (first (input-files op c)))
-         (tmp-file (process-grovel-file input-file output-file)))
-    (rename-file-overwriting-target tmp-file output-file)))
+  (destructuring-bind (output-file c-file exe-file) (output-files op c)
+    (let* ((input-file (first (input-files op c)))
+           (absolute-cache-dir (absolute-cache-dir input-file (cache-dir-of c))))
+      (process-grovel-file* input-file output-file c-file exe-file absolute-cache-dir))))
```
cbaggers: output-files has the names we computed above, so we just destructure them. 
`#'absolute-cache-dir` returns nil if the `(cache-dir-of c)` is nil

We call `process-grovel-file*` as `process-grovel-file` is the function version with the old signature

grovel/asdf.lisp
```diff
@@ -120,9 +123,10 @@
                                      :defaults input-file))
          (c-file (make-c-file-name output-file "__wrapper"))
          (o-file (make-o-file-name output-file "__wrapper"))
-         (lib-soname (wrapper-soname c)))
+         (lib-soname (wrapper-soname c))
+         (lib-file (make-lib-file-name (make-soname lib-soname output-file))))
     (list output-file
-          (make-lib-file-name (make-soname lib-soname output-file))
+          lib-file
```
cbaggers: again just some tidying

grovel/asdf.lisp
```diff
@@ -135,19 +139,37 @@
         (multiple-value-bind (files translatedp) (call-next-method)
           (values (append files (list lib-file o-file)) translatedp)))))
 
+(defun absolute-cache-dir (input-file cache-dir)
+  (when cache-dir
+    (let ((input-dir (uiop:pathname-directory-pathname input-file)))
+      (uiop:ensure-directory-pathname
+       (uiop:subpathname input-dir cache-dir)))))
+
```
cbaggers: Gets a absolute path for a directory relative to the input-file

grovel/asdf.lisp
```diff
@@ -135,19 +139,37 @@
...
+(defun system-to-component-path (component)
+  (let* ((sys (component-system component))
+         (sys-dir (pathname-directory (component-pathname sys)))
+         (comp-dir (pathname-directory (component-pathname component))))
+    (make-pathname* :directory (cons :relative (subseq comp-dir (length sys-dir))))))
+
```
cbaggers: Given an asdf component, this function returns the relative path from the system to that component

grovel/feature-caching-read.lisp
```diff
@@ -0,0 +1,78 @@
...
+
+(defun call-with-cached-reader-conditionals (func &rest args)
+  (destructuring-bind (rfunc cache) (make-caching-reader-conditional)
+    (let ((*readtable* (copy-readtable)))
+      (set-dispatch-macro-character #\# #\+ rfunc *readtable*)
+      (set-dispatch-macro-character #\# #\- rfunc *readtable*)
+      (values (apply func args) (flatten-features cache)))))
+
+(defmacro with-cached-reader-conditionals (&body body)
+  `(call-with-cached-reader-conditionals (lambda () ,@body)))
```
cbaggers: This file provides new versions of `#+` & `#-` They work exactly as before but with side effect that they also cache the [feature-expressions](http://clhs.lisp.se/Body/24_aba.htm) they were specified with.

The result of the read is a list of features pared with a bool saying whether the feature was found on the system. So something like: `((:linux t) (:windows nil))`
The result is always sorted alphabetically by feature name

grovel/grovel.lisp
```diff
@@ -80,6 +80,22 @@ int main(int argc, char**argv) {
 }
 ")
 
+(defvar *local-includes*)
```
cbaggers: This is bound by `process-wrapper-file`

grovel/grovel.lisp
```diff
@@ -179,68 +195,140 @@ int main(int argc, char**argv) {
   ;; found.
   (intern (symbol-name (car form)) '#:cffi-grovel))
 
-(defvar *header-forms* '(c include define flag typedef))
+(defvar *header-forms* '(c include-local include define flag typedef))
```
cbaggers: We need to make sure `include-local` is seen as a valid form

grovel/grovel.lisp
```diff
@@ -179,68 +195,140 @@ int main(int argc, char**argv) {
...
+        (let* ((header-forms (remove-if-not #'header-form-p forms))
+               (body-forms (remove-if #'header-form-p forms)))
+          (write-string *header* out)
+          (dolist (form header-forms)
+            (process-grovel-form out form))
+          (write-string *prologue* out)
+          (dolist (form body-forms)
+            (process-grovel-form out form))
+          (write-string *postscript* out)
+          c-file)))))
```
cbaggers: The old version was hard to read as a newcomer because of nest. It is also much smaller now as the forms are read earlier in the process and passed to this function.

grovel/grovel.lisp
```diff
@@ -179,68 +195,140 @@ int main(int argc, char**argv) {
...
+                        (mapc #'process-form (rest f)))
+                      (t (push f forms)))))
+               (process-form form)))))
+    (with-open-file (in input-file :direction :input)
+      (with-cached-reader-conditionals (read-forms in)))))
+
+(defun touch-file (pathname)
+  (with-open-file (stream pathname :direction :output
+                          :if-exists :append
+                          :if-does-not-exist :create)))
```
cbaggers: Ensures a file exists. Writes no data to it.

This is needed because of a limitation in asdf. If you tell it (in `#'perform`) that you will be making 3 files, you need to make those 3 files. There isnt a way to say 'actually we found out we didnt need them.

However this is the case when we load from cache. We don't need to create a the `c-file` as we have the `.so` file and the lisp file already. 

So we use this to make sure asdf doesnt complain to the user.

grovel/grovel.lisp
```diff
@@ -179,68 +195,140 @@ int main(int argc, char**argv) {
...
+
+(defun feature-specific-cache-dir (cache-dir feature-expressions)
+  (let ((feature-expressions (copy-seq feature-expressions)))
+    (when cache-dir
+      (ensure-directory-pathname
+       (subpathname cache-dir (gen-feature-hash feature-expressions))))))
```
cbaggers: Given the list of features we read above, generate a name for the platform specific subdirectory we will create inside the user specified `cache-dir`

grovel/grovel.lisp
```diff
@@ -179,68 +195,140 @@ int main(int argc, char**argv) {
...
+  (assert *local-includes*)
+  (let ((local-include-dir
+         (uiop:ensure-directory-pathname
+          (subpathname (uiop:pathname-directory-pathname in-dir)
+                       "local-includes"))))
+    (ensure-fresh-dir local-include-dir)
+    (loop :for (src-file . local-name) :in *local-includes*
+       :for dest-file := (subpathname local-include-dir local-name)
+       :do (alexandria:copy-file src-file dest-file))
+    (format nil "-I~A" (truename local-include-dir))))
```
cbaggers: This function is part of the `include-local` mechanism. When we discussed this feature in the mailing list, stelian said:

> I think it's better for the groveler to add the source directory to the include path, so that the distinction doesn't matter. with grovel.c being
>
> #include <foo.h>
>
> `cc -I/path/to/sources/ grovel.c` will then search for foo.c first in /path/to/sources/ and then in the default system include path.

To enable this we create a directory in the `.cache/common-lisp/etc/` cache directory and copy all the `local-include` header files into it. We later use `-I` to add this directory to the search path as suggested (which works perfectly). If you recall, in `push-local-include` we create a unique name for each `local-include` file so we know they won't clash. 

We also use `ensure-fresh-dir` to make sure any header files from previous builds are removed.

grovel/grovel.lisp
```diff
@@ -179,68 +195,140 @@ int main(int argc, char**argv) {
...
+      (let* ((cached-lisp-file (feature-specific-cache-file
+                                dest-lisp-file absolute-cache-dir
+                                feature-expressions))
+             (*local-includes* nil))
+        (if (and cached-lisp-file (file-exists-p cached-lisp-file))
+            (process-grovel-file-from-cache dest-lisp-file c-file
+                                            cached-lisp-file)
+            (process-grovel-file-from-scratch forms dest-lisp-file c-file
+                                              exe-file cached-lisp-file))
+        dest-lisp-file))))
```
cbaggers: We now have two cases, either:
- A `cache-dir` has been specified and both the files we need are inside it
OR
- The file or the directory are missing and we need to build them.

The first case we call `from-cache` and the second `from-scratch`

grovel/grovel.lisp
```diff
@@ -179,68 +195,140 @@ int main(int argc, char**argv) {
...
+    ;;
+    (handler-case (link-executable exe-file inputs)
+      (error (e) (grovel-error "~a" e)))
+    (invoke exe-file lisp-file)
+    (rename-file-overwriting-target lisp-file dest-lisp-file)
+    (when cached-lisp-file
+      (ensure-directories-exist
+       (pathname-directory-pathname cached-lisp-file))
+      (alexandria:copy-file dest-lisp-file cached-lisp-file
+                            :if-to-exists :supersede))))
```
cbaggers: Here, as in the old version, we do the bulk of the work. 
- We copy the `local-include` files to the designated directory
- We link, invoke and move the file to it's final location
- We then, if the user has specified a `cache-dir` cache the result.

grovel/grovel.lisp
```diff
@@ -723,21 +815,42 @@ string."
...
+    (loop :for c :across string :do
+       (setf hash (mod (+ (ash hash 5) hash (char-code c))
+                       wrap)))
+    hash))
+
+(defun gen-feature-hash (features)
+  (format nil "~a_~a_~x"
+          (or (operating-system) (software-type))
+          (or (architecture) (machine-type))
+          (djb2 (format nil "~{~a~}" features))))
```
cbaggers: These two functions are use when generating a name for the 'feature-specific' cache directory. 
We try to balance between human readable and terse in this implementation as:
- We dont want it to be totally readable as it would use too many characters and cause problems with windows filepath length limits
- We want the user to at least be able to see whether the OS and Architecture match their machine.
My [djb2](http://www.cse.yorku.ca/~oz/hash.html) implementation is not the prettiest but works

grovel/grovel.lisp
```diff
@@ -723,21 +815,42 @@ string."
...
+      (loop :for form = (read in nil nil) :while form :collect form))))
+
+(defun generate-c-lib-file (input-data c-file)
+  (let ((*lisp-forms* nil))
+    (ensure-directory-pathname (pathname-directory-pathname c-file))
+    (with-open-file (out c-file :direction :output
+                         :if-exists :supersede)
+      (write-string *header* out)
+      (loop :for form :in input-data
+         :do (process-wrapper-form out form)))
```
cbaggers: As with the groveller version we read with out [feature-expression](http://clhs.lisp.se/Body/24_aba.htm) caching and pass the forms to `generate-c-lib-file` once we know if we already have a result cached or not

grovel/grovel.lisp
```diff
@@ -749,12 +862,17 @@ string."
               (named-library-name
                 (let ((*package* (find-package :keyword))
                       (*read-eval* nil))
-                  (read-from-string lib-soname))))
+                  (read-from-string lib-soname)))
+              (search-file (if cached-lib
+                               `(system-relative-pathname
+                                 ,(component-name system)
+                                 ,(pathname-directory-pathname cached-lib))
+                               (directory-namestring lib-file))))
```
cbaggers: This is the part that makes sure that we can use this file portably. If we have cached the library then put the path to the library relative to the system (as it must be in the hierarchy). We use `uiop`'s `system-relative-pathname` to make sure this is portable and robust

grovel/grovel.lisp
```diff
@@ -767,22 +885,80 @@ string."
...
+             (*local-includes* nil))
+        (if (process-from-cache-p system cached-lisp-file cached-lib-file)
+            (process-wrapper-file-from-cache system dest-lib-file dest-lisp-file
+                                             c-file o-file
+                                             cached-lib-file cached-lisp-file)
+            (process-wrapper-file-from-scratch system input-data dest-lisp-file
+                                               lib-soname dest-lib-file c-file
+                                               o-file cached-lib-file
+                                               cached-lisp-file))
+        (values dest-lisp-file dest-lib-file)))))
```
cbaggers: As with the groveller version higher up in this file, this needs to check if we have the files cached so it knows if it needs to build them ('from-scratch') or if it can just copy them ('from-cache')

grovel/grovel.lisp
```diff
@@ -767,22 +885,80 @@ string."
...
+        (delete-file tmp-file))
+      (when (and cached-lisp-file cached-lib-file)
+        (let ((cached-lisp-file (system-relative-pathname system cached-lisp-file))
+              (cached-lib-file (system-relative-pathname system cached-lib-file)))
+          (ensure-directories-exist
+           (pathname-directory-pathname cached-lib-file))
+          (alexandria:copy-file dest-lisp-file cached-lisp-file
+                                :if-to-exists :supersede)
+          (alexandria:copy-file lib-file cached-lib-file
+                                :if-to-exists :supersede))))))
```
cbaggers: These two are very similar to the groveller versions.

grovel/grovel.lisp
```diff
@@ -824,6 +1000,10 @@ string."
 (define-wrapper-syntax include (&rest includes)
   (format out "~{#include <~A>~%~}" includes))
 
+(define-wrapper-syntax include-local (&rest paths)
+  (format out "~{#include <~A>~%~}"
+          (mapcar #'push-local-include paths)))
+
```
cbaggers: The same as the grovel version

PR #107, opened 2016-10-15
Null name shouldnt satisfy strong rules

cbaggers:
The problem with this that with forms like this:

    typedef enum {
      dJointTypeNone = 0,
      dJointTypeBall,
      dJointTypeHinge,
    } dJointType;

The enum is always excluded, as the name is nil. Then the `dJointType`
typedef tries too lookup the anonomous entry for the enum and fails.

The only way to counter it is to have explicitly include the definitions
for everything in the files (which makes it feel much less automatic).

This commit is one possible fix. Another is to allow the user to
customize the `include-definition?` predicate. Also `cl-autowrap`
handles this case so maybe we can take some inspiration from there.

The worrying thing of course is that this is a potentially breaking change

### DIFF
--- a/src/c2ffi/generator.lisp
+++ b/src/c2ffi/generator.lisp
@@ -123,12 +123,12 @@
   (labels
       ((covered-by-a-rule? (name rules)
          (or (eq rules :all)
-             (null name)
              (not (null (some (rcurry #'funcall name) rules)))))
        (weak? (rules)
          (eq :all rules))
        (strong? (name rules)
-         (and (not (weak? rules))
+         (and name
+              (not (weak? rules))
               (covered-by-a-rule? name rules))))
     (let* ((excl-def/weak   (weak? exclude-definitions))
            (excl-def/strong (strong? name exclude-definitions))


cbaggers:
As an example I can now use the following to allow all definitions in the header files specified but none in any others:

```
              :include-sources ("collision.h"
                                "collision_space.h"
                                "collision_trimesh.h"
                                "common.h"
                                "contact.h"
                                "error.h"
                                "mass.h"
                                "objects.h"
                                "ode.h"
                                "odeinit.h")
              :exclude-sources :all```

attila-lendvai:
re breaking change: it doesn't matter, i think we are the only two users of this code for now... :)

do i understand it right that with the patch applied there will always be generated an anonymous typedef for your example? not too nice, but i don't have immediate ideas how to easily avoid that... and maybe it's not such a big issue?

cbaggers:
> re breaking change: it doesn't matter, i think we are the only two users of this code for now... :)

hehe ok

> do i understand it right that with the patch applied there will always be generated an anonymous typedef for your
> example?

Yeah right now that's what happens as it is two separate json entries in the `.spec` file. I agree that it would be nice to reduce this when we can but that feels like a separate issue.

So yeah, not such a big issue for now.

attila-lendvai:
i wrote a proposal/RFC to cffi-devel regarding this: https://mailman.common-lisp.net/pipermail/cffi-devel/2016-October/003007.html

cbaggers:
Cool it would be good to have some simpler mechanism for this.

Back on this issue though, I think this still should be merged as the idea of something satisfying a 'strong' rule just by having a null name seems wrong. I think a strong rule being something explicit is a good idea and the rest of the code follows this

attila-lendvai:
ok, i'll look into merging this regardless.

re cffi:defctype syntax: if it understood inline declarations (just like how C deftype understand an inline enum), then we could just record every anonymous entry in the c2ffi output without emitting it, and emit them in place where referenced (they are connected through an internal ID).

i'll also look into c2ffi, maybe it could be fixed there.

cbaggers:
> we could just record every anonymous entry in the c2ffi output without emitting it, and emit them in place where referenced
Good idea. I clearly didnt have enough coffee in my system earlier :)

Thanks for merging

attila-lendvai:
i added support for this into CFFI, but the problem is that the inline enum is not a toplevel macro, so it cannot emit the `defparameter`s that `defcenum` emits... i've never been sure about those `defparamter`s, and i don't know how to proceed from here.

they are useful, but maybe there should be just an API for emitting them and users should be explicit about it. an API to enumerate enum's and to emit `defconstant`s for them. i'll think about this and open a new issue if i have something more substantial to share.

attila-lendvai:
so, this will not work, c2ffi cannot trivially implement the required change, plus there are C level issues. for the details see the linked c2ffi issue above.

i think i'll put this on the shelf. with your patch merged it's only an aesthetic issue now.

PR #108, opened 2016-10-15
*read-default-float-format* should be double-float

cbaggers:
Otherwise it crashes the first time it tries to read one. The example
case for me was in:

    #define DBL_EPSILON __DBL_EPSILON__

### DIFF
--- a/src/c2ffi/generator.lisp
+++ b/src/c2ffi/generator.lisp
@@ -543,6 +543,7 @@ target package."
                (*ffi-name-transformer* (canonicalize-transformer-hook ffi-name-transformer))
                (*ffi-type-transformer* (canonicalize-transformer-hook ffi-type-transformer))
                (*ffi-export-predicate* (canonicalize-transformer-hook ffi-export-predicate))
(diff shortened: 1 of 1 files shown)


attila-lendvai:
merged, thanks!

sionescu:
I just saw this now. IMO this should be reverted because the standard says the default is 'SINGLE-FLOAT: http://clhs.lisp.se/Body/v_rd_def.htm
If CFFI breaks somehow with the standard default it should be made to work that way.

attila-lendvai:
@sionescu, why is the CLHS relevant in this context?

i don't see how the higher precision float can cause anything unwanted here, but i'm open for getting corrected. keep in mind that this is only about the cffi/c2ffi generator code, not cffi as a whole.

cbaggers:
Just to add clarification this change was to ensure cl-json doesn't simply crash when it tries to read a literal double.

I would agree that, when we emit and :float constant declaration from as lisp code, that we should coerce to a single-float, however that is a separate issue and still require cl-json to be able to read doubles. PR welcome.

PR #110, opened 2016-12-18
Propagate search-path when searching for alternatives.

borodust:
When `:search-path` is specified in library definition and `load-foreign-library` called, if cffi searches for `:or` alternatives, it won't find those in the provided `:search-path`, because these paths are not propagated into search function.

### DIFF
--- a/src/libraries.lisp
+++ b/src/libraries.lisp
@@ -327,12 +327,12 @@ ourselves."
               (report-simple-error name error)))
           (report-simple-error name error))))))
 
-(defun try-foreign-library-alternatives (name library-list)
+(defun try-foreign-library-alternatives (name library-list &optional search-path)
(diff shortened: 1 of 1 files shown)


borodust:
@attila-lendvai is there anything specific I should add to PR? Tests, comments, or mb I should rephrase the description?

attila-lendvai:
sorry, but i don't understand this part well enough, let alone the multi platform aspects of it, so i don't feel the authority to commit this. i'm only one of the maintainers.

attila-lendvai:
although, this looks rather straightforward...

@luismbo, if that helps, then if you give me the green light, i can deal with the administrative part, if any.

luismbo:
@borodust, @attila-lendvai thanks!

PR #111, opened 2016-12-25
Fix dependencies in cffi-toolchain.asd

fare:
The dependencies were subtly wrong: when the ASDF loaded was already > 3.1.6,
then the asdf-compat was blackholed by its if-feature condition,
which broke the chain of dependencies for package.


luismbo:
Heh, subtle indeed. Thanks!

PR #112, opened 2017-01-23
Improve c-toolchain

fare:
Add support for getting parameters from new SBCL sbcl.mk.
Add support for getting parameters from CLISP makevars.
Add support for getting parameters for ECL and MKCL.

Refactor static linking to use the above.
On SBCL, this relies on the new feature :sb-linkable-runtime.
On ECL and MKCL, have static-{image,program}-op be pass-through synonyms
for the usual variants.
Remove deceitful incomplete support for static linking on CMUCL.

Add shell test for static linking, to better work with SBCL (uses cl-launch).

Add class c-file for C files to compile, and o-file for precompiled objects.
They can be either dynamically loaded into the current process,
or statically linked into a new runtime.

Upgrade backward compatibility bundle from asdf 3.1.6 to 3.2.0.
Stop manually curating a minimal set of changes,
just adopt a new bundle.lisp wholesale.


fare:
This branch is stable enough, but probably in need of review from CFFI maintainers.

It depends on changes to SBCL that are not upstream yet, but I believe that dougk will be merging my changes within the next few weeks.

fare:
Xof just merged my patch to upstream SBCL, so the interface this uses is stable.

cffi-toolchain.asd
```diff
@@ -25,6 +25,8 @@
 ;;; DEALINGS IN THE SOFTWARE.
 ;;;
 
+;; Load before we even read the #.(if ...) below.
```
attila-lendvai: "load"? this LOAD-SYSTEM is here to force an upgrade, no?

attila-lendvai:
if no one speaks up, i'll eventually push this into master.

cffi-toolchain.asd
fare: Yes.

fare: Updated the comment.

attila-lendvai:
merged, thanks!

PR #113, opened 2017-02-21
Fix dynamic linking with SB-LINKABLE-RUNTIME

fare:
Shared objects were failing to link on the latest SBCL because it specifies
-fno-pie in its CFLAGS, but that invalidates the previously specified -fPIC.
Fix that by specifying -fPIC *after* the SBCL-provided CFLAGS, not *before*.

Also, remove a redundant use of -fPIC, and fix a comment.


luismbo:
Thanks Faré.

PR #114, opened 2017-03-16
Evaluate :count, :size, :offset forms while creating 'defcstruct'.

ANDRON94:
Current  implementation of *defcstruct* doesn't allows to put a valid Lisp form to a :size, :count, :offset options. For example, for a next form:
```lisp
(cffi:defcstruct (test-struct :size (* 20 5))
   (one :uint32 :count (* 3 4))
   (two :double :offset (* 4 5)))
```
There are next results:
- compiler throws an error: "type of (* 3 4) is not a REAL" or "type of (* 4 5) is not a NUMBER";
- and '(cffi:foreign-type-size '(:struct test-struct))' returns a "(* 20 5)".

I think, such type of numeric fields should be evaluated cause it's pretty useful to set, for example, *:count* option as "(* +my-constant+ +my-other-constant+)"

### DIFF
--- a/grovel/asdf.lisp
+++ b/grovel/asdf.lisp
@@ -146,8 +146,7 @@
            (alexandria:copy-file tmp-file output-file :if-to-exists :supersede)
         (delete-file tmp-file))))
 
-
-;; Allow for naked :grovel-file and :wrapper-file in asdf definitions.
+;; Allow for naked :cffi-grovel-file and :cffi-wrapper-file in asdf definitions.
 (setf (find-class 'asdf::cffi-grovel-file) (find-class 'grovel-file))
 (setf (find-class 'asdf::cffi-wrapper-file) (find-class 'wrapper-file))
 
--- a/toolchain/c-toolchain.lisp
+++ b/toolchain/c-toolchain.lisp
@@ -248,7 +248,7 @@ is bound to a temporary file name, then atomically renaming that temporary file 
 
 (defun cc-compile (output-file inputs)
   (apply 'invoke-builder (list *cc* "-o") output-file
-         "-c" #-windows "-fPIC" (append *cc-flags* inputs)))
+         "-c" (append *cc-flags* #-windows '("-fPIC") inputs)))
 
 (defun link-executable (output-file inputs)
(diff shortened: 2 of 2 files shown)


liamh:
#. should solve the problem, e.g. use #.(* 20 5)

ANDRON94:
@liamh Yes, I've already use it as workaround =)
But #. is evaluating while *read*, so for "(defparameter +my-const+ 45)" it is not working and I should to write something like that:
```lisp
(eval-when (:compile-toplevel :load-toplevel :execute)
   (defparameter +my-const+ 45))

....
#.(* +my-const+ 4)
```

guicho271828:
that should be a defconstant

ANDRON94:
Yeah, *defconstant* and *#.()* is a quite good combination. And I understand that it isn't the most useful feature but anyway in such kind of options I expect to use any computable form(constants, functions, etc) without *#.()-magic* =)
Should I discard this request?

guicho271828:
Apologies for my reflex comment, I didn't think much about the topic itself.
> Should I discard this request?
No, I think it is a good feature. Few coments:

+ adding a typecheck would help detecting the error early.
+ the docstring of defstruct should note that those forms are evaluated

guicho271828:
can you cherry-pick this? https://github.com/guicho271828/cffi

guicho271828:
I mean 8ed6abf

ANDRON94:
@guicho271828 Yeah, sure. Maybe I need to look more deeply into eval question in cffi librrary. I think, tomorrow or the day after tomorrow my pull request will be ready.

luismbo:
We should avoid `eval` though, otherwise we lose the lexical scope.

ANDRON94:
@luismbo Yes, you're right. Forms should be expanded in the lexical environment. But there is no expansion at all, only a function call.

guicho271828:
ping

ANDRON94:
@guicho271828 Hi, I should totally rework this PR because of @luismbo note.
Unfortunately, I have no much time for a next few days. So, probably I will start working on it next week.

PR #116, opened 2017-03-17
migrate from cl-travis to roswell

guicho271828:
* cl-travis depends on an unmaintained library CIM
* allow `sudo : false` for faster builds
* allow builds on OSX
* cmucl, abcl, ecl and others
* stop manually fetching external libraries

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,36 +1,42 @@
-branches:
-  only:
-    - master
+language: common-lisp
+sudo: false
 
-language: lisp
+addons:
+  apt:
+    packages:
+      - libc6-i386
+      - clisp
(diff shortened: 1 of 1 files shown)


guicho271828:
CIM is officially unmaintained http://keens.github.io/blog/2017/01/29/deprecating_cim/

luismbo:
Grabbing dependencies via git is a way to allow testing of bleeding edge features and catching incompatibilities before quicklisp releases. Unless you have a good reason to remove that feature, I'd rather keep that.

guicho271828:
thanks for the advice. I reverted that change

guicho271828:
I should re-work on this.

PR #122, opened 2017-05-04
Robustify static linking command printing

fare:
Make the static linking command more robust.


luismbo:
Care to elaborate on what exactly has been robustified? :-)

attila-lendvai:
but why? isn't he's the sole user? :)

fare:
I updated the commit with additional comments and reformatting and a better commit message.

It is right that I am maybe the only user at this point `:-(`

The original edit was actually done from quite a few weeks ago. I believe I reconstituted the motivation for it in the commit message. Anyway, I was flushing this change from my local repository after a friend asked me a question about cffi-grovel and I wanted to get my code in synch with his.

luismbo:
We need Google Analytics embedded into ASDF or something to track the number of users!

Anyhoo, thanks @fare.

PR #123, opened 2017-05-31
Adds link to launchpad for bug reporting

vydd:


guicho271828:
ping @attila-lendvai

attila-lendvai:
it's not my responsibility to decide. and if it were, i would go with the github issue tracker.

@luismbo may have something to add here.

vydd:
To give some context now that 9 months have passed, I did this because someone on #lisp told me that  all bugs should go to launchpad, so I thought that that info should be available in the README as well.

fiddlerwoaroof:
There is no issue tracker on the github repository and no link to Launchpad, which makes it hard to see what issues have been reported.  Can we merge this commit in?

luismbo:
Fixed via c28c760. I guess we need to replicate the website info in the README.md since that's probably got much more visibility these days.

PR #125, opened 2017-10-10
Make static linking more resilient to ASDF upgrade

fare:
Configure and upgrade ASDF before loading quicklisp.


luismbo:
Thanks @fare.

PR #126, opened 2017-11-03
grovel: don't lose if CFLAGS is set.

stassats:
Split the string into a list and don't discard other *cc-flags*
parameters.


luismbo:
@stassats cffi-grovel really needs a proper test suite... :-/

PR #128, opened 2018-01-02
Add &REST MAKE-ARRAY-ARGS to #'FOREIGN-ARRAY-TO-LISP

phoe:
Rename test ARRAY.ROUND-TRIP to ARRAY.FOREIGN-TO-LISP.BASIC
Add test ARRAY.FOREIGN-TO-LISP.ADJUSTABLE
Add test ARRAY.FOREIGN-TO-LISP.DISPLACED
Add test ARRAY.FOREIGN-TO-LISP.SPECIALIZED

Implementation detail for test ARRAY.FOREIGN-TO-LISP.SPECIALIZED:

15.1.2.2 of the standard states that the only truly portable array specializations are for bits (bit-vectors) and characters (strings). Since char-codes are implementation-dependent, it would be tricky to write a portable test for them without generating characters at runtime. So, for a truly portable test, we are only left with bits, which are luckily numeric, and equal to (UNSIGNED-BYTE 1).

This is why this test is so terribly wasteful, spending a whole byte for a single bit - CFFI has no capabilities for dealing with single bits, and this test is only meant to check correctness of the :ELEMENT-TYPE argument to
MAKE-ARRAY. In actual use cases of specialized FOREIGN-ARRAY-TO-LISP, capable implementations will be able to make specialized arrays of types that are commonly optimized for and/or representable in hardware, such as (UNSIGNED-BYTE 8) on x86 architectures.

### DIFF
--- a/src/types.lisp
+++ b/src/types.lisp
@@ -388,7 +388,8 @@ newly allocated memory."
       form))
 
 (defun lisp-array-to-foreign (array pointer array-type)
-  "Copy elements from a Lisp array to POINTER."
+  "Copy elements from a Lisp array to POINTER. ARRAY-TYPE must be a CFFI array
+type."
   (let* ((type (ensure-parsed-base-type array-type))
          (el-type (element-type type))
          (dimensions (dimensions type)))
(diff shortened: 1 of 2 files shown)


luismbo:
https://bugs.launchpad.net/cffi/+bug/1740800

luismbo:
Thanks @phoe!

phoe:
@luismbo <3

PR #129, opened 2018-02-20
translate non-keyword foreign enum keys properly

3b:
%foreign-enum-value accepts any non-NIL symbol, so don't limit it to
KEYWORDP in translate-to-foreign and expand-to-foreign.

Also fixes style-warning in SBCL when passing keywords to inlined
foreign function calls with SPEED 3.

### DIFF
--- a/src/enum.lisp
+++ b/src/enum.lisp
@@ -219,7 +219,7 @@
         (%foreign-enum-keyword type-obj value :errorp errorp))))
 
 (defmethod translate-to-foreign (value (type foreign-enum))
-  (if (keywordp value)
+  (if (symbolp value)
       (%foreign-enum-value type value :errorp t)
       value))
 
@@ -233,7 +233,7 @@
 
 (defmethod expand-to-foreign (value (type foreign-enum))
   (once-only (value)
-    `(if (keywordp ,value)
+    `(if (symbolp ,value)
          (%foreign-enum-value ,type ,value :errorp t)
          ,value)))
 


3b:
style warning part is https://bugs.launchpad.net/cffi/+bug/1750647, but KEYWORDP seems incorrect in general

luismbo:
Hey @attila-lendvai, ecc5e24e8e2328ce9c1fb37271c19e23b00e641b is what made `defcenum` accept non-keyword symbols as keys. I have a vague recollection that this may have been motivated by cffi/c2ffi. Is that accurate? Can you remind me why?

attila-lendvai:
@luismbo i don't remember all the details, but c2ffi generates a CL package and uses only that for the C namespace. i think it's much cleaner to deal with bluez:some-c-enum-member than to litter the global keyword namespace with stuff from the C namspece brought over by an automated tool.

src/enum.lisp
```diff
@@ -219,7 +219,7 @@
         (%foreign-enum-keyword type-obj value :errorp errorp))))
 
 (defmethod translate-to-foreign (value (type foreign-enum))
-  (if (keywordp value)
+  (if (symbolp value)
```
attila-lendvai: there's a deftype called enum-key for this

attila-lendvai:
@luismbo re your lp comment: then we should map all of the C namespace to one CL package, which is a bad idea.

there's no obvious isomorphism between the C and CL namespace handling, so users want to have the freedom to do their own mappings in their own CL packages. well, at least this user... :)

luismbo:
Closed by

PR #130, opened 2018-04-27
Use sb-rt in sbcl instead of rt package

kierangrant:
SBCL provides a version of rt called sb-rt.


luismbo:
Thanks for the pull request but I think it's best to use the same library across all Lisps.

PR #131, opened 2018-06-06
Evaluate :count, :size, :offset forms while creating 'defcstruct'.

ANDRON94:
Hello, I've finally completed this feature. It required more job than I expected.
Changelog:
- refactor 'defcstruct' and 'defcunion' macros. So now :size, :count, and :offset options are expanded in the lexical environment of macro definition;
- add more unit tests to cover new feature;
- describe the new feature in the documentation;

The old PR - #114

### DIFF
--- a/src/utils.lisp
+++ b/src/utils.lisp
@@ -27,11 +27,12 @@
 
 (in-package #:cffi)
 
-(defmacro discard-docstring (body-var &optional force)
-  "Discards the first element of the list in body-var if it's a
-string and the only element (or if FORCE is T)."
-  `(when (and (stringp (car ,body-var)) (or ,force (cdr ,body-var)))
-     (pop ,body-var)))
+(defun omit-docstring (body-var)
+  "Return the original elements of BODY-VAR list but
+without the first element if the type of this element is the string."
+  (if (stringp (first body-var))
+      (rest body-var)
+      body-var))
 
 (defun single-bit-p (integer)
   "Answer whether INTEGER, which must be an integer, is a single
--- a/doc/cffi-manual.texinfo
+++ b/doc/cffi-manual.texinfo
@@ -2744,14 +2744,16 @@ CFFI> (foreign-type-size 'foo)
 @node defcunion, defctype, defcstruct, Foreign Types
 @heading defcunion
 @subheading Syntax
-@Macro{defcunion name &body doc-and-slots @res{} name}
+@Macro{defcunion name-and-options &body doc-and-slots @res{} name}
+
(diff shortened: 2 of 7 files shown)


src/utils.lisp
```diff
@@ -27,11 +27,12 @@
...
-  "Discards the first element of the list in body-var if it's a
-string and the only element (or if FORCE is T)."
-  `(when (and (stringp (car ,body-var)) (or ,force (cdr ,body-var)))
-     (pop ,body-var)))
+(defun omit-docstring (body-var)
+  "Return the original elements of BODY-VAR list but
+without the first element if the type of this element is the string."
+  (if (stringp (first body-var))
+      (rest body-var)
+      body-var))
```
luismbo: `discard-docstring` doesn't destroy anything, it simply points `body-var` elsewhere.

dkochmanski: also if body has only one element, then string is returned value, not a docstring (at least in Common Lisp), so function name should be omit-car-string or something in this spirit

```
(defun xxx () "HI!") ;-> XXX
(defun yyy () "HI!" "BYE") ;-> YYY
(xxx) ;-> "HI!")
(yyy) ;-> "BYE"
(documentation 'xxx 'function) ;-> NIL
(documentation 'yyy 'function) ;-> "HI!"
```

ANDRON94: > discard-docstring doesn't destroy anything

Why?  'discard-docstring' expands into (pop ,body-var) and from Hyperspec:
> The effect of (pop place) is roughly equivalent to
>  (prog1 (car place) (setf place (cdr place)))

So, body-var is modified which is forbidden for macro arguments.

> function name should be omit-car-string or something in this spirit

Yeah, I agree with you, maybe "omit-first-string"?

luismbo: `(let ((x (list 1 2))) (setf x (rest x)))` doesn't modify the list `(1 2)`, it points `x` to the second cons cell. The first cons cell remains intact.

sionescu: I would use parse-declarations:parse-body instead of coming with your own (https://common-lisp.net/project/parse-declarations/manual/parse-declarations.html#parse_002dbody).

ANDRON94: @luismbo Ok, I understand my mistake. So, in the end, I've just changed the signature of the function(change return value from void to list).
I think, that @dkochmanski is right and 'omit-first-string' is a better name.
@sionescu I only changed the function that was used before. I don't sure that we need to use a separate library for this.
What are significant problems with this commit? Is it useless?

PR #132, opened 2018-09-24
Add groveller syntax to detect if preprocessor constants are defined

sdilts:
This adds a new groveller form `feature` that detects if a preprocessor constant is defined, and adds a value to `*features*` if the constant is present. This is intended to be used when a C library defines preprocessor constants if it was compiled with certain functionality.

### DIFF
--- a/doc/cffi-manual.texinfo
+++ b/doc/cffi-manual.texinfo
@@ -6685,6 +6685,13 @@ constant will be a symbol of the same name interned in the current
 package.
 @end deffn
 
+@deffn {Grovel Form} feature lisp-feature-name c-name &key feature-list
+
+Adds @var{lisp-feature-name} to the list @var{feature-list} if the @var{c-name}
+string is known to the C preprocessor. @var{feature-list} defaults
+to @code{cl:*features*}.
+@end deffn
+
 @deffn {Grovel Form} define name &optional value
 
 Defines an additional C preprocessor symbol, which is useful for
--- a/grovel/grovel.lisp
+++ b/grovel/grovel.lisp
@@ -354,6 +354,16 @@ int main(int argc, char**argv) {
   (dotimes (i (length c-names))
     (format out "~&#endif~%")))
 
+(define-grovel-syntax feature (lisp-feature-name c-name &key (feature-list 'cl:*features*))
+  (c-section-header out "feature" lisp-feature-name)
+  (format out "~&#ifdef ~A~%" c-name)
+  (c-format out "(cl:pushnew '")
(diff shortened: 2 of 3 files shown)


doc/cffi-manual.texinfo
```diff
@@ -6685,6 +6685,12 @@ constant will be a symbol of the same name interned in the current
 package.
 @end deffn
 
+@deffn {Grovel Form} feature lisp-feature-name c-name
+
+Adds lisp-feature-name to @var{*features*} if c-name is known to the C
+preprocessor.
```
luismbo: This should be

```
Adds @var{lisp-feature-name} to @code{*features*} if @var{c-name} is known to the C preprocessor.
```

grovel/grovel.lisp
```diff
@@ -354,6 +354,12 @@ int main(int argc, char**argv) {
   (dotimes (i (length c-names))
     (format out "~&#endif~%")))
 
+(define-grovel-syntax feature (lisp-feature-name c-name)
+  (c-section-header out "feature" lisp-feature-name)
+  (format out "~&#ifdef ~A~%" c-name)
+  (c-format out "(cl:pushnew ~S cl:*features*)" lisp-feature-name)
+  (format out "~&#endif~%"))
```
luismbo: `cl:*features*` is quite convenient for `#+` and `#-`, but for some use cases perhaps a boolean variable/predicate would work better or an alternative list variable (with `*features*` as the default)? Did you consider any of these alternatives?

sdilts: Both of those suggestions sound like a good idea, especially the option to use a different lists. Both would be easy enough to add. How about splitting this into two forms?
``` lisp
;; for the list variant
(feature-list lisp-feature-name c-name &key (feature-list cl:*features*))
;; for the variable
(feature-var bool-name c-name)
```

luismbo: I think I'd be happy with just the first variant. Because features are fun to develop, hard to maintain and impossible to remove... 🙂 Having said that, perhaps we can fold the second variant into the existing `constant` clause?

sdilts:
I added the option to add to a different list and cleaned up the documentation in the manual.

luismbo:
@sdilts I tried to write a test and I think we were mishandling non-keyword symbols. This is a symptom of cffi-grovel's funky handling of packages, but using `c-print-symbol` for the feature name seems like the right thing to do?

sdilts:
It looks like `c-print-symbol` is the way to go; the test shows it produces the  correct output, and cursory tests show the same. I don't know what else we would use.

PR #133, opened 2018-10-17
Fix bug where flags included with :cc-flags in asd file have no effect

sdilts:
The c files are generated during the `process-op` operation, not the `compile-op` operation. This fixes a bug where specifying compiler flags `:cc-flags` in the asdf file has no effect.

To reproduce the error, attempt to grovel a file and use `:cc-flags` to specify the include path. For example, place the following files in the same directory and attempt to load the `test` system.

`test.h`:
``` c
struct test {
   int a;
};
```
`grovel-file.lisp`:
``` lisp
(include "test.h")
```
`test.asd`:
``` lisp
(asdf:defsystem #:test-grovel
  :defsystem-depends-on (#:cffi-grovel)
  ;; replace #.(...) with the path to this file's directory if 
  ;; (namestring *load-pathname*) doesn't return this file's directory
  :components ((:cffi-grovel-file "grovel-file" 
                                  :cc-flags #.(concatenate 'string "-I" (directory-namestring *load-pathname*)))))
```
The system will not load, and you will see that the additional compiler flag was not added.


luismbo:
@fare have you got a chance to review this ASDF-related change?

fare:
LGTM.

I admit this is all out of my L2 cache, but the explanation seems quite convincing, and I see nothing obviously wrong with the patch.

@luismbo thanks for tagging me — I hadn't otherwise been notified, though I am responsible for this bug indeed. Mea culpa and thanks @sdilts for the fix!

luismbo:
@fare oh, I tagged for your ASDF expertise. 🙂 Thanks!

PR #136, opened 2018-11-09
Push static-image-op feature into child lisp environment

fiddlerwoaroof:
This is a fairly minor change that adds a feature flag to the child lisp process to indicate that we're in the process of building a static image.


luismbo:
Tell me more about how you then use this flag and when it should be
cleared.

fiddlerwoaroof:
Maybe there's a better way to do it, but I used this to prevent osicat from generating a .so and depending on it.  Perhaps it would be better to redefine the wrapper-file component in the child environment?

https://gitlab.com/fiddlerwoaroof/osicat/commit/b56f924c5dafa92e6da1f2a91a6b7f133cbd69fa

fiddlerwoaroof:
As far as when it should be cleared goes, I suppose the right time to do that would be right before saving the child environment?

PR #137, opened 2018-12-23
check for ENUM-KEY instead of KEYWORD in foreign enum translators

3b:
KEYWORDP can't be constant folded, since a symbol could be uninterned
after compilation, so using a more general type gets rid of some
style warnings on SBCL. %FOREIGN-ENUM-VALUE has an explicit check for
ENUM-KEY, so that is probably more correct than KEYWORD anyway.

### DIFF
--- a/src/enum.lisp
+++ b/src/enum.lisp
@@ -219,7 +219,7 @@
         (%foreign-enum-keyword type-obj value :errorp errorp))))
 
 (defmethod translate-to-foreign (value (type foreign-enum))
-  (if (keywordp value)
+  (if (typep value 'enum-key)
       (%foreign-enum-value type value :errorp t)
       value))
 
@@ -233,7 +233,7 @@
 
 (defmethod expand-to-foreign (value (type foreign-enum))
   (once-only (value)
-    `(if (keywordp ,value)
+    `(if (typep ,value 'enum-key)
          (%foreign-enum-value ,type ,value :errorp t)
          ,value)))
 


3b:
without it, code like ` (gl:draw-arrays :triangles 0 6))` gets style warnings like

```

;     (CL-OPENGL-BINDINGS:DRAW-ARRAYS :TRIANGLES 0 6)
; --> BLOCK MULTIPLE-VALUE-PROG1 CFFI:FOREIGN-FUNCALL LET LET IF
; ==>
;   #:VALUE6
;
; caught STYLE-WARNING:
;   This is not a (UNSIGNED-BYTE 32):
;    :TRIANGLES
;   See also:
;     The SBCL Manual, Node "Handling of Types"
```
`draw-arrays` is defined by something that expands to

```lisp
(PROGN
 (DECLAIM (INLINE DRAW-ARRAYS))
 (DEFUN DRAW-ARRAYS (MODE FIRST COUNT)
   (MULTIPLE-VALUE-PROG1
       (LET ((#:G560
              (LET ((#:VALUE566 MODE))
                (IF (KEYWORDP #:VALUE566)
                    (CFFI::%FOREIGN-ENUM-VALUE #<CFFI::FOREIGN-ENUM ENUM> #:VALUE566
                                               :ERRORP T)
                    #:VALUE566))))
         (LET ((#:G561 (TRUNCATE FIRST)))
           (LET ((#:G562 (TRUNCATE COUNT)))
             (CFFI-SYS:%FOREIGN-FUNCALL "glDrawArrays"
                                        (:UNSIGNED-INT #:G560 :INT #:G561 :INT #:G562
                                         :VOID)
                                        :CONVENTION :CDECL :LIBRARY OPENGL))))
     (CHECK-ERROR 'DRAW-ARRAYS))))
```

3b:
which seems to be a dup of #129, though at least with the predefined type this time :/

luismbo:
So, my only objection was that the original intent of defcenum was to use keywords and that's what's documented in the manual. I'll fix the manual then.

PR #138, opened 2019-01-17
Add functions to add and remove features during a static image op

daewok:
STATIC-IMAGE-NEW-FEATURES must return a list a features to be added in the
static runtime.

STATIC-IMAGE-REMOVE-FEATURES-ON-DUMP must return a list of features that will be
removed as part of UIOP's image dump hook.

### DIFF
--- a/examples/main-example.lisp
+++ b/examples/main-example.lisp
@@ -11,6 +11,10 @@
 (defun entry-point ()
   (when uiop:*command-line-arguments*
     (uiop:format! t "Arguments: ~A~%" (uiop:escape-command uiop:*command-line-arguments*)))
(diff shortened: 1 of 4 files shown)


daewok:
Similar to #136 I want to push some new features into the static runtime. My use case is that I'm statically linking the SBCL runtime with libssl. To prevent cl+ssl from trying to load the libraries itself, I need to add `:cl+ssl-foreign-libs-already-loaded` to the child lisp. I have no need to remove that feature before dumping, but figured a way to do that should be added anyways.

I think this is complementary, but orthogonal, to #136 as there's probably still a use case for having a feature that's unconditionally added and removed.

PR #139, opened 2019-02-28
Add MacPorts lib/ to darwin-fallback-library-path

jrjsmrtn:
`darwin-fallback-library-path` doesn't support the MacPorts library path `/opt/local/lib`, causing issues like https://github.com/dimitri/pgloader/issues/161 and https://github.com/dimitri/pgloader/issues/444

This patch adds the MacPorts library path before `/usr/local/lib` and `/usr/lib`.

I successfully tested it on MacOS X 10.14.3, MacPorts 2.5.4, SBCL 1.5.0 and pgloader 3.6.4ec8613.


luismbo:
Why before /usr/local/lib and /usr/lib?

luismbo:
I guess they're sorted from most to least advanced/tweaked. Makes sense. Thanks!

PR #140, opened 2019-03-02
define with-pointer-to-vector-data, correct make-shareable-byte-vector

kpoeck:
For implementation clasp
* Define missing with-pointer-to-vector-data
* make-shareable-byte-vector

Implementation now possible that clasp supports static-vectors.

Tested with cl+ffi

### DIFF
--- a/src/cffi-clasp.lisp
+++ b/src/cffi-clasp.lisp
@@ -137,13 +137,13 @@ SIZE-VAR is supplied, it will be bound to SIZE during BODY."
 (defun make-shareable-byte-vector (size)
   "Create a Lisp vector of SIZE bytes that can passed to
 WITH-POINTER-TO-VECTOR-DATA."
-  (make-array size :element-type '(unsigned-byte 8)))
-
-;; frgo, 2016-07-02: TODO: Implemenent!
-;; (defmacro with-pointer-to-vector-data ((ptr-var vector) &body body)
-;;   "Bind PTR-VAR to a foreign pointer to the data in VECTOR."
-;;   `(let ((,ptr-var (si:make-foreign-data-from-array ,vector)))
-;;      ,@body))
+  ;; borrowed from static.vectors
+  (core:make-static-vector '#.(upgraded-array-element-type '(unsigned-byte 8)) size))
+  
+;; borrowed from static.vectors
+(defmacro with-pointer-to-vector-data ((ptr-var vector) &body body)
+  `(let ((,ptr-var (core:static-vector-pointer ,vector 0)))
+     ,@body))
 
 (defun %foreign-type-size (type-keyword)
   "Return the size in bytes of a foreign type."


src/cffi-clasp.lisp
```diff
@@ -137,13 +137,13 @@ SIZE-VAR is supplied, it will be bound to SIZE during BODY."
...
   "Create a Lisp vector of SIZE bytes that can passed to
 WITH-POINTER-TO-VECTOR-DATA."
-  (make-array size :element-type '(unsigned-byte 8)))
-
-;; frgo, 2016-07-02: TODO: Implemenent!
-;; (defmacro with-pointer-to-vector-data ((ptr-var vector) &body body)
-;;   "Bind PTR-VAR to a foreign pointer to the data in VECTOR."
-;;   `(let ((,ptr-var (si:make-foreign-data-from-array ,vector)))
-;;      ,@body))
+  ;; borrowed from static.vectors
```
luismbo: What's `static.vectors`?

kpoeck: See https://github.com/clasp-developers/static-vectors for clasp or
https://github.com/sionescu/static-vectors

Basically this allows an array containing e.g. simple bytes to be passed to ffi-world (with the guarantee that the gc won't move the memory).
With-pointer-to-vector-data and make-shareable-byte-vector do very similar things, so I borrowed from clasp implementations of static-vectors

luismbo: Oh, OK. I think you can drop those comments. @sionescu won't mind. :-)

src/cffi-clasp.lisp
```diff
@@ -137,13 +137,13 @@ SIZE-VAR is supplied, it will be bound to SIZE during BODY."
...
+  (core:make-static-vector '#.(upgraded-array-element-type '(unsigned-byte 8)) size))
```
luismbo: I'm unsure about this call to upgraded-array-element-type. What's it doing? Is it really possibly that some clasp platforms don't support ub8 arrays? Would it signal an error if you passed '(unsigned-byte 8) unconditionally? /cc @sionescu

kpoeck: Yes, errors;
(core:make-static-vector  '(unsigned-byte 8) 12) 
->
Condition of type: SIMPLE-PROGRAM-ERROR
Handle make-static-vector :element-type (UNSIGNED-BYTE 8)

(upgraded-array-element-type '(unsigned-byte 8)) 
->
EXT:BYTE8

The code is borrowed from here (from drmeister, main author of clasp)
https://github.com/clasp-developers/static-vectors/blob/master/src/impl-clasp.lisp#L22

kpoeck:
@luismbo , I can certainly drop the comments you mentionned, but I don't fully understand what I should do regarding the  upgraded-array-element-type
I can substitute with 'ext:byte8, but do you consider that better than the current code

luismbo:
I'm not sure whether calling `upgraded-array-element-type` or using `ext:byte8` is better. The latter seems clearer to me, but the former is probably more future-proof. Your call.

src/cffi-clasp.lisp
luismbo: https://github.com/cffi/cffi/pull/140#issuecomment-471514272

sionescu:
I think it would be better to add a dependency on static-vectors and wrap its primitives instead of cut-and-paste.

luismbo:
The shareable byte vector API has slightly different constraints and requirements. In particular, it doesn't require the vectors to be static. The implementation can use pinning instead. (Some less sophisticated implementations even fallback to copying.) Also, the vector must be GCable. So, I think using static-vectors might be a good idea, but it would plenty of work outside of the scope of @kpoeck's improvement.

PR #142, opened 2019-04-28
Make the library work with libffi.so.7

vaartis:
This makes CFFI work with libffi `3.3_rc0`+, as `ABI_SYSV` was removed there. I'm not sure if it'd continue to work with lower versions, but it should, since `ABI_UNIX64` was there all along.

### DIFF
--- a/libffi/libffi-types.lisp
+++ b/libffi/libffi-types.lisp
@@ -71,6 +71,7 @@
 #-(or freebsd windows)
 (cenum abi
  ((:default-abi "FFI_DEFAULT_ABI"))
+ #-x86-64
  ((:sysv "FFI_SYSV"))
  ((:unix64 "FFI_UNIX64")))
 
--- a/libffi/libffi.lisp
+++ b/libffi/libffi.lisp
@@ -31,8 +31,8 @@
   (:darwin (:or "libffi.dylib" "libffi32.dylib" "/usr/lib/libffi.dylib"))
   (:solaris (:or "/usr/lib/amd64/libffi.so" "/usr/lib/libffi.so"))
   (:openbsd "libffi.so")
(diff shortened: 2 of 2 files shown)


luismbo:
I'm not sure where those constant names come from. I can't find them in libffi's source code. Shouldn't they be `FFI_SYSV`, `FFI_UNIX64`, etc?

vaartis:
I personally took this value from [gentoo wiki article on migrating from an older version](https://wiki.gentoo.org/wiki/Libffi_3.3_porting_notes/FFI_SYSV) (it has a commit reference to the change)

luismbo:
@vaartis thanks for the link. It looks like `FFI_SYSV` was only removed on 64-bit platforms. Does this alternative patch work for you?

```diff
--- a/libffi/libffi-types.lisp
+++ b/libffi/libffi-types.lisp
@@ -71,6 +71,7 @@
 #-(or freebsd windows)
 (cenum abi
  ((:default-abi "FFI_DEFAULT_ABI"))
+ #-x86-64
  ((:sysv "FFI_SYSV"))
  ((:unix64 "FFI_UNIX64")))
```

vaartis:
It seems to.

luismbo:
@vaartis cool, can you update the pull request accordingly? BTW, I've noticed that we're not propagating abi info from `foreign-funcall` to `libffi` so I've registered that bug on launchpad: https://bugs.launchpad.net/cffi/+bug/1826884

luismbo:
@vaartis I think you've accidentally dropped the foreign library definition changes (adding `"libffi.so.7"` and `"libffi32.so.7"`). While you're at it, maybe add `libffi-7.dll` to the `:windows` clause too.

luismbo:
Thanks @vaartis!

PR #145, opened 2019-06-30
Improved with-foreign-slots binding abilities to allow local names for slots

stacksmith:
Now accepts with-slots-like bindings (name slotname), as well as (name :pointer slotname).  All in all a binding may be one of:
```
  name                       name bound to slot of same name              
  (:pointer name)            name bound to pointer to slot of same name
  (name slot-name)           name bound to slot-name
  (name :pointer slot-name)  name bound to pointer to slot-name
```
Rationale:   

Lispers expect binding forms to provide the ability to bind chosen names.  The inability of `with-foreign-slots`to do so is both surprising and disappointing.  It makes it awkward to use `with-foreign-slots` in functions that happen to bind symbols that happen to coincide with slot names:
```
(defun point-init (point x y)
   (with-foreign-slots ((x y) point  (:struct point))
       (setf x... ---Yikes!---
```
It is also nice to use names you like instead of slot names:
```
(with-foreign-slots (((w width)(h height)) rect (:struct rect))
  (let ((area (* w h)))
 ...
```
When using structures from another package, it is now possible to create slot accessors locally, and save a lot of package noise:
```
(with-foreign-slots (((x gtk::x)(y gtk::y)) point (:struct gtk::point))
  (+ x 11) ; x and y are now local accessors.
...
```

### DIFF
--- a/src/types.lisp
+++ b/src/types.lisp
@@ -835,7 +835,7 @@ The foreign array must be freed with foreign-array-free."
       (foreign-struct-slot-set-form
        value ptr (get-slot-info (eval type) (eval slot-name)))
       form))
-
+#||
 (defmacro with-foreign-slots ((vars ptr type) &body body)
   "Create local symbol macros for each var in VARS to reference
 foreign slots in PTR of TYPE. Similar to WITH-SLOTS.
@@ -855,6 +855,30 @@ case slot-name will be bound to the pointer to that slot."
                        "Malformed slot specification ~a; must be:`name' or `(:pointer name)'"
                        var))
                   `(,var (foreign-slot-value ,ptr-var ',type ',var))))
+         ,@body))))
+||#
+(defmacro with-foreign-slots ((vars ptr type) &body body)
+  "Create local symbol macros for each var in VARS to reference
+foreign slots in PTR of TYPE. Similar to WITH-SLOTS.
+Each var can be of the form: 
+  name                       name bound to slot of same name              
+  (:pointer name)            name bound to pointer to slot of same name
+  (name slot-name)           name bound to slot-name
+  (name :pointer slot-name)  name bound to pointer to slot-name"
+
+  (let ((ptr-var (gensym "PTR")))
+    `(let ((,ptr-var ,ptr))
+       (symbol-macrolet
+           ,(loop :for var :in vars
+	       :collect
+		 (if (listp var)
+		     (let ((p1 (first var)) (p2 (second var)) (p3 (third var)))
+		       (if (eq p1 :pointer)	
+			   `(p2 (foreign-slot-pointer ,ptr-var ',type ',p2))
+			   (if (eq p2 :pointer)
+			       `(p1 (foreign-slot-pointer ,ptr-var ',type ',p3))
+			       `(p1 (foreign-slot-value ,ptr-var ',type ',p2)))))
+		     `(,var (foreign-slot-value ,ptr-var ',type ',var))))
          ,@body))))
 
 ;;; We could add an option to define a struct instead of a class, in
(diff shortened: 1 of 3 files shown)


luismbo:
Looks great! How about extracting the pointer functionality to a second macro? `with-foreign-slot-pointers` or something. Also, please update the manual?

stacksmith:
You can already use pointers.  No changes needed. 
```
(with-foreign-slots (((ptrx :pointer gtk::x)) point (:struct gtk::point))
   (mem-ref ptrx :int))

**Comment by luismbo** [comment]
Heh. Oh well, I don't particularly like the pointer syntax, but it's probably too late to fix it. :-)

**Comment by stacksmith** [comment]
Well, it's really fine.  It's better than a separate with-foreign-slot-pointers macro, for sure as it allows any mix of slot values or pointers, even both for the same slot.

**Comment by stacksmith** [comment]
Make a note: the test suite does not verify the (with-foreign-slots) :pointer syntax.  Tests came up as correct even though the :pointer parser was broken...

**Comment by luismbo** [comment]
Yes, I don't think there are tests for the pointer syntax, nor for the new name syntax. Can you add some?

**Comment by stacksmith** [comment]
Is there any reason this was never merged?

File: src/types.lisp [hunk 1]
```diff
@@ -856,6 +856,30 @@ case slot-name will be bound to the pointer to that slot."
...
+||#
+(defmacro with-foreign-slots ((vars ptr type) &body body)
+  "Create local symbol macros for each var in VARS to reference
+foreign slots in PTR of TYPE. Similar to WITH-SLOTS.
+Each var can be of the form: 
+  name                       name bound to slot of same name              
+  (:pointer name)            name bound to pointer to slot of same name
+  (name slot-name)           name bound to slot-name
+  (name :pointer slot-name)  name bound to pointer to slot-name"
+
```
luismbo, line 868: Remove this empty line.

tests/struct.lisp
```diff
@@ -703,3 +703,30 @@
   42
   (1 . 2)
   42)
```
luismbo, line 705: Missing empty line here.

luismbo (changes requested):
Sorry, this got buried under a big pile of email. Just have few very, very minor requests.

stacksmith:
luismbo, thanks for picking it back up. I made the requested changes.  Please let me know if there is anything else.

src/types.lisp
```diff
@@ -839,22 +839,24 @@ The foreign array must be freed with foreign-array-free."
...
-                  `(,var (foreign-slot-value ,ptr-var ',type ',var))))
+	       :collect
+		 (if (listp var)
+		     (let ((p1 (first var)) (p2 (second var)) (p3 (third var)))
+		       (if (eq p1 :pointer)	
+			   `(,p2 (foreign-slot-pointer ,ptr-var ',type ',p2))
+			   (if (eq p2 :pointer)
+			       `(,p1 (foreign-slot-pointer ,ptr-var ',type ',p3))
+			       `(,p1 (foreign-slot-value ,ptr-var ',type ',p2)))))
+		     `(,var (foreign-slot-value ,ptr-var ',type ',var))))
```
luismbo, line 859: Oh no, tabs! Did you use a non-Emacs editor to write this? 😱

stacksmith: Sorry, I attempted to make the fixes right in the browser...   Let me try again.

stacksmith: OK, I think I fixed it.

luismbo (approved):

luismbo:
It’s a bit annoying that Travis is not working at the moment. Might be a good time to move to GitHub pages as SLIME did.

luismbo:
Thanks @stacksmith.

stacksmith:
Thank you.

PR #146, opened 2019-07-18
Respect CFLAGS env var and support MacPorts include directory

orivej:
Users should be able to pass custom compilation flags such as include directories via `CFLAGS` env var.

It would be convenient for MacPorts users if libraries installed in `/opt/local` worked by default. (`/opt/local/lib` is supported since #139.)

Fixes https://github.com/orivej/pzmq/pull/18


toolchain/c-toolchain.lisp
```diff
@@ -223,12 +220,21 @@
           *linkkit-start* nil
           *linkkit-end* nil)))
 
+(defun common-toolchain-parameters ()
+  (setf *cc-flags*
+        (append
+         *cc-flags*
+         ;; For MacPorts
+         #+darwin (list "-I" "/opt/local/include/")
```
yitzchak: If you change this to `#+(darwin openbsd freebsd) (list "-I" "/opt/local/include/")` then you can eliminate the freebsd include above (line 214) and also then lines 35-42 in [libffi/libffi-types.lisp](https://github.com/cffi/cffi/blob/86836561ffd173cfa7a9ea2d9bd26dd6aa49b614/libffi/libffi-types.lisp#L35)

Just a thought.

orivej: Thanks, done. (Used `/usr/local/include/` for BSD.)

orivej:
@luismbo, would you merge this?

toolchain/c-toolchain.lisp
```diff
@@ -223,12 +218,23 @@
...
+         #+(or freebsd openbsd) (list "-I" "/usr/local/include/"))))
+
 (defun ensure-toolchain-parameters ()
   #+clisp (unless *cc* (clisp-toolchain-parameters))
   #+ecl (unless *cc* (ecl-toolchain-parameters))
   #+mkcl (unless *cc* (mkcl-toolchain-parameters))
   #+sbcl (unless *cc* (sbcl-toolchain-parameters))
-  (unless *cc* (default-toolchain-parameters)))
+  (unless *cc* (default-toolchain-parameters))
+  (common-toolchain-parameters))
```
luismbo: What's going on here? In which case is `*cc*` null? i.e., why do we need `common-toolchain-parameters`?

luismbo:
I'm having trouble reading this commit, sorry. Which bits are "respecting CFLAGS"?

toolchain/c-toolchain.lisp
orivej: We need `common-toolchain-parameters` because apart from toolchain-specific parameters (such as SBCL-specific parameters based on its `sbcl.mk`) we want to have some common and OS-specific parameters: (1) `*cc-flags*` based on the `CFLAGS` environment variable, (2) `/opt/local/include` on Darwin.

The case of null `*cc*` was handled by `default-toolchain-parameters` before this PR. This PR augments `*cc-flags*` after `*cc-flags*` and `*cc*` were set.

orivej:
> Which bits are "respecting CFLAGS"?

This line: https://github.com/cffi/cffi/pull/146/commits/77ba274cddebacffc4de7a6a3e723bfccc14ea59#diff-55403a5ad4f9128e5a1f36a78166dd05R224
(Without this PR CFFI on SBCL ignores CFLAGS.)

toolchain/c-toolchain.lisp
luismbo: Is it that *cc* is being checked to see if ensure-toolchain-parameters has run before?

If so, is it problematic that the new function you've added doesn't check for it?

luismbo:
I'm even more confused now. 🤔 You've just added that commit. Does that mean it wasn't respecting CFLAGS before that latest commit?

toolchain/c-toolchain.lisp
orivej: Indeed, there is a surprising second call of ensure-toolchain-parameters here: https://github.com/cffi/cffi/blob/v0.20.1/toolchain/static-link.lisp#L6 . I have moved the body into `unless *cc*`, https://github.com/cffi/cffi/pull/146/commits/5c1f1f67 . (Alternatively the call in `static-link.lisp` may be deleted and `unless *cc*` removed.)

orivej:
> You've just added that commit. Does that mean it wasn't respecting CFLAGS before that latest commit?

The main use case of CFLAGS is to let user specify `-I` paths to the non-default locations where they have installed the headers of their libraries. In this case it is irrelevant whether CFLAGS overrides toolchain defaults or not, but in general it seems better to let CFLAGS overrides toolchain defaults, so I moved them above toolchain-specific `*cflags*`.

toolchain/c-toolchain.lisp
```diff
@@ -223,12 +218,24 @@
...
-  #+ecl (unless *cc* (ecl-toolchain-parameters))
-  #+mkcl (unless *cc* (mkcl-toolchain-parameters))
-  #+sbcl (unless *cc* (sbcl-toolchain-parameters))
-  (unless *cc* (default-toolchain-parameters)))
+  (unless *cc*
+    #+clisp (clisp-toolchain-parameters)
+    #+ecl (ecl-toolchain-parameters)
+    #+mkcl (mkcl-toolchain-parameters)
+    #+sbcl (sbcl-toolchain-parameters)
+    (unless *cc* (default-toolchain-parameters))
```
orivej: Note that `(sbcl-toolchain-parameters)` does not set `*cc*` on Travis. Here is the check where `(unless *cc* (default-toolchain-parameters))` was replaced by `#-(or clisp ecl mkcl sbcl) (default-toolchain-parameters)`: https://travis-ci.org/cffi/cffi/jobs/575204326

yitzchak:
@luismbo Any chance on this one? Should resolve my yitzchak/common-lisp-jupyter#32 and make grovel work more easily on MacPorts.

notmgsk:
`b u m p`

PR #149, opened 2019-11-16
on ABCL add :default keyword default arg to %foreign-funcall

slyrus:
* this allows the ABCL varargs tests to pass


luismbo:
Merged, thanks.

PR #152, opened 2020-02-21
fix translation of struct-typedefs

enometh:
CFFI now always translates aggregate struct values to lisp property
lists.  Earlier struct values which were defined by types introduced
by defctype were translated as foreign pointers.

One consequence of this change is that defcvar to a foreign structure
will always produce an immutable lisp list. Users will have to resort
to FOREIGN-SYMBOL-VALUE in order to access structs to modify them -
and use the (:POINTER VAR) syntax of WITH-FOREIGN-SLOTS to access and
modify fields within the struct.

* src/early-types.lisp: (defctype): use translatable-foreign-type
instead of enhanced-foreign-type.

* tests/structs.lisp: New tests: (misnamed) STRUCT-VALUES.FSBV.1
STRUCT-VALUES.FSBV.1 Updated tests: STRUCT.ALIGNMENT.1
STRUCT.ALIGNMENT.2 STRUCT.ALIGNMENT.3 STRUCT.ALIGNMENT.4
STRUCT.ALIGNMENT.5 STRUCT.ALIGNMENT.6 STRUCT.ALIGNMENT.7
STRUCT.NESTED-SETF STRUCT.ALIGNMENT.8

### DIFF
--- a/tests/struct.lisp
+++ b/tests/struct.lisp
@@ -122,9 +122,9 @@
 
 (deftest struct.alignment.1
     (list 'a-char (foreign-slot-value
-                   (foreign-slot-pointer *the-s-s-ch* 's-s-ch 'a-s-ch)
+                   (foreign-slot-pointer (get-var-pointer '*the-s-s-ch*) 's-s-ch 'a-s-ch)
                    's-ch 'a-char)
-          'another-char (foreign-slot-value *the-s-s-ch* 's-s-ch 'another-char))
+          'another-char (foreign-slot-value (get-var-pointer '*the-s-s-ch*) 's-s-ch 'another-char))
   (a-char 1 another-char 2))
 
 
@@ -144,7 +144,7 @@
 (defcvar "the_s_s_short" s-s-short)
 
 (deftest struct.alignment.2
-    (with-foreign-slots ((yet-another-char a-s-short) *the-s-s-short* s-s-short)
+    (with-foreign-slots ((yet-another-char (:pointer a-s-short)) (get-var-pointer '*the-s-s-short*) s-s-short)
       (with-foreign-slots ((a-char another-char a-short) a-s-short s-short)
         (list 'a-char           a-char
               'another-char     another-char
@@ -171,7 +171,7 @@
 
 (deftest struct.alignment.3
     (with-foreign-slots
-        ((yet-another-char a-s-double a-short) *the-s-s-double* s-s-double)
+        ((yet-another-char (:pointer a-s-double) a-short) (get-var-pointer '*the-s-s-double*) s-s-double)
       (with-foreign-slots ((a-char a-double another-char) a-s-double s-double)
         (list 'a-char            a-char
               'a-double          a-double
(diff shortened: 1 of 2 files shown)


tests/struct.lisp
```diff
@@ -118,7 +118,7 @@
 
 (defctype s-s-ch (:struct s-s-ch))
 
-(defcvar "the_s_s_ch" s-s-ch)
+(defvar *the-s-s-ch* (foreign-symbol-pointer "the_s_s_ch"))
```
luismbo: Nice idea.

This has the drawback of not working with saved cores since it'll permanently store that pointer. Not a big deal for the test suite, but since this code might be used as an example, maybe change `defvar` do `define-symbol-macro`.

Alternatives:

1. Keep the `defcvar` and sprinkle `get-var-ptr` throughout.
2. Change the `defctype`s to something that doesn't translate structs, either with a custom type or by adding some option to the `:struct` type that skips translation. Dubious. Probably too much work.

I'm fine with your proposed solution (with `defvar` changed to `define-symbol-macro`).

tests/struct.lisp
```diff
@@ -141,10 +141,10 @@
 
 (defctype s-s-short (:struct s-s-short))
 
-(defcvar "the_s_s_short" s-s-short)
+(defvar  *the-s-s-short* (foreign-symbol-pointer "the_s_s_short"))
```
luismbo: Extra space there between `defvar` and the variable name. 🕵️‍♂️

luismbo:
Thanks!

luismbo:
I'll have to revert this as it's causing issues with magicl:

https://github.com/rigetti/magicl/issues/106

defcfun expansion before: https://pastebin.com/tjeyKh39 (see lines labeled ; *****)
defcfun expansion after: https://pastebin.com/X6eGV4ga

test case: https://pastebin.com/7pLMVcFA

luismbo:
https://bugs.launchpad.net/cffi/+bug/1924764

enometh:
Luis, check your mail - on June 20 2020 I sent you the fix to cffi  libffi which would would have made this pull request work - correctly

The original problem is with old code in cffi libffi and not a bug in this patch which was reverted.

I wrote:
> I've pushed the fix for review on my personal branch as
> https://github.com/enometh/cffi/commit/74a15af823e3bfec7256e9a572c67962ed985f60.patch
Again mentioned on cffi-devel on Nov 06 2020 in a reply to Steve Nunez.

Unfortunately there was no feedback from him

I had hoped you would test this patch against the launchpad problems noted above - i believe it would indeed solve them 


vert could be reverted.

[translate-objects-ret-fix-translatable-foreign.patch.txt](https://github.com/cffi/cffi/files/6329774/translate-objects-ret-fix-translatable-foreign.patch.txt)

luismbo:
I see, sorry. Emails have this nasty habit of piling up and disappearing. I've pushed the stuff you sent me by email to this pull request: https://github.com/cffi/cffi/pull/181 Did I miss anything?

PR #153, opened 2020-02-23
Fixed c-toolchain.lisp to work on OpenBSD.

tapioco71:


toolchain/c-toolchain.lisp
```diff
@@ -173,8 +173,9 @@
              (if normalizep (normalize-flags sbcl-home val) val))))
     (unless (featurep :sb-linkable-runtime)
       (setf *linkkit-start* nil *linkkit-end* nil))
+    #+openbsd
+    (setf *cc-flags* (list "-I" "/usr/local/include/"))
     (setf *ld* *cc* ;; !
-          #+openbsd *cc-flags* (list "-I" "/usr/local/include/")
```
attila-lendvai: random note: the thinko fixed here demonstrates nicely why, even though it's possible, multi-entry setf's are a bad idea...

toolchain/c-toolchain.lisp
```diff
@@ -173,6 +173,8 @@
              (if normalizep (normalize-flags sbcl-home val) val))))
     (unless (featurep :sb-linkable-runtime)
       (setf *linkkit-start* nil *linkkit-end* nil))
+    #+openbsd
+    (setf *cc-flags* (list "-I" "/usr/local/include/"))
```
luismbo: Is this bit necessary? This portion is SBCL-only.

toolchain/c-toolchain.lisp
luismbo: Also, I think reader conditionals are the wrong tool here, particularly for CLISP that has portable fasls, but CFFI's fasls are not portable anyway. :-(

PR #154, opened 2020-04-04
Enum field value references

egao1980:
C standard allows references to other enumeartion fields:
> Semantics
> 3 The identifiers in an enumerator list are declared as constants that have type int and
> may appear wherever such are permitted.109) An enumerator with = defines its
> enumeration constant as the value of the constant expression. If the first enumerator has
> no =, the value of its enumeration constant is 0. Each subsequent enumerator with no =
> defines its enumeration constant as the value of the constant expression obtained by
> adding 1 to the value of the previous enumeration constant. (The use of enumerators with
> = may produce enumeration constants with values that duplicate other values in the same
> enumeration.) The enumerators of an enumeration are also known as its members.

Example from ICU4C:
```c
/** Possible types of UCalendars 
 * @stable ICU 2.0
 */ 
enum UCalendarType {
  /**
   * Despite the name, UCAL_TRADITIONAL designates the locale's default calendar,
   * which may be the Gregorian calendar or some other calendar.
   * @stable ICU 2.0
   */
  UCAL_TRADITIONAL,
  /**
   * A better name for UCAL_TRADITIONAL.
   * @stable ICU 4.2
   */
  UCAL_DEFAULT = UCAL_TRADITIONAL,
  /**
   * Unambiguously designates the Gregorian calendar for the locale.
   * @stable ICU 2.0
   */
  UCAL_GREGORIAN
};
```

This is a trivial change that brings this to CFFI.

Proposed CFFI declaration would be:
```lisp
(defcenum UCalendarType
  :UCAL_TRADITIONAL
  (:UCAL_DEFAULT :UCAL_TRADITIONAL)
  :UCAL_GREGORIAN)
```


phantomics:
Thanks Nikolai, I second merging this PR, it would help get more libraries working through CFFI.

src/enum.lisp
```diff
@@ -78,6 +78,16 @@
...
+        ;; resolve field reference
+        (when (keywordp value)
+          (setf value (gethash value keyword-values)))
+        (when (listp value)
+          (setf value (eval
+                       (list 'let (loop for k
+                                        being the hash-key
+                                        using (hash-value v) of keyword-values
+                                        collect (list (intern (string k)) v))
+                             value))))
```
luismbo, line 90: What's going on here in this `(when (listp value) ...)` branch?

doc/cffi-manual.texinfo
```diff
@@ -2871,11 +2872,15 @@ CFFI> (foreign-enum-value 'boolean :no)
 @lisp
 (defcenum numbers
   (:one 1)
+  (:another_one :one)
```
luismbo, line 2875: Underscores are very unlispy. Maybe use `:another-one` instead?

tests/enum.lisp
```diff
@@ -229,3 +229,52 @@
...
+  :year_woy 
+  :dow_local
+  :extended_year
+  :julian_day
+  :milliseconds_in_day
+  :is_leap_month
+  :field_count
+  (:day_of_month :date))
+
+(deftest enum.keyword-ref.2
```
luismbo, line 273: Such a long defcenum example. Is this testing anything different from the previous test?

luismbo (changes requested):
Sorry, this flew under my radar!

tests/enum.lisp
egao1980: It's just a real world example I had issues with. I'll review those.

.travis.yml
```diff
@@ -1,6 +1,7 @@
 branches:
   only:
     - master
+    - enum-field-references
```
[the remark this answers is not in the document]
egao1980: Sure

src/enum.lisp
egao1980: This evaluates the value of enum field in cases like ```(:UDAT_FULL_RELATIVE (cl:logior UDAT_FULL UDAT_RELATIVE))```.
The code creates (let ...) binding current values of :UDAT_FULL and :UDAT_RELATIVE to variables UDAT_FULL and  UDAT_RELATIVE correspondingly and to evaluate (cl:logior UDAT_FULL UDAT_RELATIVE).

luismbo: I see. Did you consider computing these derived values in SWIG when generating the bindings? These seems like implementation details leaking into the binding. Nevertheless, I can see how this might be convenient even for writing bindings by hand.

This approach is problematic though, as field names can collide with the names of Lisp constants in which case evaluating that `let` form will yield an error.

AeroNotix:
Ref: https://github.com/cffi/cffi/pull/311

egao1980:
Please revisit this one. Really need it for SWIG wrappers in tesseract, ICU4C and others...

PR #155, opened 2020-05-04
Implement an enum flag that elides an error when translating an enum value that was not declared.

Shinmera:
This is useful for very large enumerations of which we're only interested in a subset, or enumerations that allow for arbitrary client or vendor extensions that we cannot be aware of. Erroring on a value that we ostensibly do not care about in such a situation is not desirable, and instead simply passing through the integer allows handling the interesting cases easily in a CASE or something similar.

If there's any style inconsistencies or other changes I should make, please let me know.

### DIFF
--- a/doc/cffi-manual.texinfo
+++ b/doc/cffi-manual.texinfo
@@ -2821,7 +2821,7 @@ you should use @ref{define-foreign-type}.
 @Macro{defcenum name-and-options &body enum-list}
 
 enum-list ::= [docstring] @{ keyword | (keyword value) @}*
-name-and-options ::= name | (name &optional (base-type :int))
+name-and-options ::= name | (name &optional (base-type :int) &key allow-undeclared-values)
 
 @subheading Arguments and Values
 
@@ -2834,6 +2834,10 @@ A documentation string, ignored.
 
 @item base-type
 A symbol denoting a foreign type.
+
+@item allow-undeclared-values
+Whether to pass through integer values that were not explicitly declared
+in the enum when translating from foreign memory.
 
 @item keyword
 A keyword symbol.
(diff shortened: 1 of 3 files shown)


src/enum.lisp
```diff
@@ -55,7 +55,11 @@
    (value-keywords
     :initform (error "Must specify VALUE-KEYWORDS.")
     :initarg :value-keywords
-    :reader value-keywords))
+    :reader value-keywords)
+   (undeclared-values-allowed-p
+    :initform NIL
```
luismbo, line 60:
```suggestion
    :initform nil
```

doc/cffi-manual.texinfo
```diff
@@ -2855,6 +2859,14 @@ being passed as arguments to or returned from foreign functions,
 respectively. The same applies to any other situations where an object
 of an @code{enum} type is expected.
 
+If a value should be translated to lisp that is not declared in the
+enum, an error will be signalled. You can elide this error and instead
+make it pass the original enum value by specifying
+@var{undeclared-values-allowed-p}. This can be useful for very large
```
luismbo, line 2865: minor style suggestion: maybe rename this to `allow-undeclared-values` à la `allow-other-keys`?

tests/enum.lisp
```diff
@@ -130,13 +130,30 @@
...
+  3)
+
 ;;;# Bitfield tests
 
 ;;; Regression test: defbitfield was misbehaving when the first value
 ;;; was provided.
 (deftest bitfield.1
     (eval '(defbitfield (bf1 :long)
-             (:foo 0)))
+            (:foo 0)))
```
luismbo, line 156: Revert this? Or maybe change the quote to backquote to get proper automatic indentation?

doc/cffi-manual.texinfo
Shinmera: I wasn't sure whether the 'predicates should be -p things' conventions should be followed or not. I'll change it to what you suggested.

Shinmera:
Okey, your changes should be addressed now.

doc/cffi-manual.texinfo
luismbo: That convention usually applies to function names, not argument names, I think.

tests/enum.lisp
```diff
@@ -130,6 +130,23 @@
   3.42
   4.42)
 
+;; Test undeclared values
+(defcenum (enum.undeclared :int :undeclared-values-allowed-p T)
```
luismbo, line 134:
```suggestion
(defcenum (enum.undeclared :int :allow-undeclared-values t)
```

PR #156, opened 2020-05-13
toolchain: remove shared objects when dumping a static image.

ralt:
Otherwise, sbcl tries to load them again when restoring the
image/program, which is pointless given that they were compiled within
the dumped image. (And is actually causing errors if the library
doesn't exist at runtime.)

### DIFF
--- a/toolchain/static-link.lisp
+++ b/toolchain/static-link.lisp
@@ -33,6 +33,12 @@
 
 ;; Problem? Its output may conflict with the program-op output :-/
 
+#+sbcl
+(defmethod perform :before ((op static-image-op) (s system))
+  (loop for object in sb-alien::*shared-objects*
+     do (setf (sb-alien::shared-object-dont-save object) t))
+  (setf *image-entry-point* (ensure-function (asdf/system:component-entry-point s))))
+
 #-(or ecl mkcl)
 (defmethod perform ((o static-image-op) (s system))
   #-(or clisp sbcl) (error "Not implemented yet")


ralt:
cc @stassats because it's using sbcl unexported symbols, but I can't think of a better way there.

stassats:
I guess you gotta use internal symbols for now, as I'm not ready to commit to a stable API.

ralt:
Ok, thanks for the quick feedback.

sionescu:
The problem here is that you can't guarantee that sb-alien::*shared-objects* only contain libraries that are "seen" by cffi-toolchain and thus statically linked. The failure mode will be seeing "undefined alien" errors (IIRC) or segfaults when starting the saved image.

ralt:
@sionescu do you have any example of that? Maybe we can only do this for those that cffi-toolchain has seen, and warn for the other found shared objects. But I'd need a reproducer to play with.

stassats:
I gathered if you want a static image the you don't want any libraries. But I don't really know what a     static image is.

ralt:
At least with a warning I'd be able to know that asdf/cffi could not do their job correctly for a given library, and react accordingly.
As far as I understand, a static image is essentially:

- Build all the shared libraries into a .a static library
- Build an image using that .a + the sbcl.o

ralt:
FWIW, I've tried building various libraries:

- osicat (relying on built-by-asdf shared lib)
- cl-sqlite (relying on system-provided shared lib)

And the libraries were embedded correctly every time. So I'd like to get an example of a shared library seen by sbcl but not by cffi-toolchain, because I can't find any.

luismbo:
@ralt does cffi-toolchain link sqlite statically?

ralt:
I was going to say yes, but my testing was wrong. A binary using sqlite does start fine (e.g. `--help`), but gets alien errors when trying to use the library.

I'm closing this issue until I've figured out if we can statically linked those instead. Thanks!

PR #157, opened 2020-05-15
toolchain: remove non-system libaries in STATIC-IMAGE-OP

ralt:
Which applies to STATIC-PROGRAM-OP as well, because it's a
subclass. Essentially, the non-system libraries have been statically
linked (C-wise) into the image, so we don't need to have them as
foreign libraries. In fact, as foreign libraries, the restored image
will try to load them and fail, hence defeating the point.

System libraries do need to be kept, though, as they're shared
libraries, and not embedded in the image at all.

Caveat: this has only been tested on sbcl so far.

### DIFF
--- a/toolchain/static-link.lisp
+++ b/toolchain/static-link.lisp
@@ -33,6 +33,16 @@
 
 ;; Problem? Its output may conflict with the program-op output :-/
 
+(defmethod perform :before ((op static-image-op) (s system))
+  ;; Close non-system foreign libraries, as grovel ones are embedded
+  ;; in the static-image. System foreign libraries still need to be
+  ;; explicitly loaded on startup, so we're keeping them loaded.
+  (register-image-dump-hook
+   (lambda ()
+     (dolist (library (list-foreign-libraries))
+       (when (eql (foreign-library-type library) :grovel-wrapper)
+	 (close-foreign-library library))))))
+
 #-(or ecl mkcl)
 (defmethod perform ((o static-image-op) (s system))
   #-(or clisp sbcl) (error "Not implemented yet")
--- a/toolchain/package.lisp
+++ b/toolchain/package.lisp
@@ -24,7 +24,7 @@
 ;;;
 
 (uiop:define-package #:cffi-toolchain
-  (:mix #:asdf #:uiop #:common-lisp)
+  (:mix #:asdf #:uiop #:cffi #:common-lisp)
   (:import-from #:asdf/bundle
    #:link-op #:bundle-pathname-type #:bundle-type
    #:gather-operation #:gather-type)


toolchain/static-link.lisp
```diff
@@ -33,6 +33,18 @@
...
 
+(defmethod perform :before ((op static-image-op) (s system))
+  ;; Close non-system foreign libraries, as grovel ones are embedded
+  ;; in the static-image. System foreign libraries still need to be
+  ;; explicitly loaded on startup, so we're keeping them loaded.
+  (register-image-dump-hook
+   (lambda ()
+     (loop
+	for library in (list-foreign-libraries)
+	unless (eq (foreign-library-type library) :system)
```
sionescu, line 44: Use EQL and invert the test: check if the type is :GROVEL-WRAPPER.

ralt: Done.

toolchain/static-link.lisp
```diff
@@ -33,6 +33,18 @@
...
 ;; Problem? Its output may conflict with the program-op output :-/
 
+(defmethod perform :before ((op static-image-op) (s system))
+  ;; Close non-system foreign libraries, as grovel ones are embedded
+  ;; in the static-image. System foreign libraries still need to be
+  ;; explicitly loaded on startup, so we're keeping them loaded.
+  (register-image-dump-hook
+   (lambda ()
+     (loop
+        for library in (list-foreign-libraries)
```
luismbo, line 43: Minor style suggestion, reformat as

```
(loop for library in (list-foreign-libraries)
      when (eql ...)
        do (close ...)) 
```

do you have slime-indentation enabled?

Or just use dolist, that'd work too.

toolchain/static-link.lisp
```diff
@@ -33,6 +33,18 @@
...
+        when (eql (foreign-library-type library) :grovel-wrapper)
+        do (close-foreign-library library))))
+  (setf *image-entry-point* (ensure-function (asdf/system:component-entry-point s))))
```
luismbo, line 46: What does this line do?

toolchain/static-link.lisp
ralt: dolist sounds good, using that.

toolchain/static-link.lisp
ralt: Given that I'm overriding `defmethod perform :before` from `program-op` (see `bundle.lisp`), I need to re-do it. It's most likely lack of Lisp knowledge from my part, as I tried calling `(call-next-method)`, unsuccessfully.

toolchain/static-link.lisp
ralt: Pushed an update with that.

toolchain/static-link.lisp
luismbo: All applicable :before methods will run in most-specific-first order, so there's no overriding going on.

ralt: Oh, I'm redefining the PERFORM method on STATIC-IMAGE-OP instead of STATIC-PROGRAM-OP...

ralt: Fixed now.

toolchain/static-link.lisp
```diff
@@ -33,6 +33,16 @@
 
 ;; Problem? Its output may conflict with the program-op output :-/
 
+(defmethod perform :before ((op static-program-op) (s system))
```
luismbo, line 36: static-program-op inherits from static-image-op. IIUC, your fix is useful for both.

ralt: Done.

sionescu:
The more I look at the toolchain code, the more I think it should just be deleted. The code is impenetrable and too tied up to ASDF, where it belongs.
My 2 cents.

ralt:
As far as I understand, you and @luismbo are the CFFI maintainers. Can you take a decision on whether you're going to remove cffi-toolchain altogether or not? Because if yes, there's not much point in me spending time on it.

luismbo:
We'll keep it for the foreseeable the future. Even if we were to find a better home for it, it wouldn't be dropped, it'd just move.

ralt:
Thanks for the clarification.

PR #160, opened 2020-05-18
Detect statically linked libraries and avoid loading

sionescu:
* add option :CANARY to DEFINE-FOREIGN-LIBRARY: a foreign symbol unique
to the library

If that symbol is found in the current image before
the first attempt to load the library, the latter is assumed to be
statically linked and LOAD-FOREIGN-LIBRARY only marks the load state
as :STATIC before returning.

* add new slot LOAD-STATE which, together with HANDLE, tracks the
library

| library state     | LOAD-STATE | HANDLE  |
|-------------------|------------|---------|
| not loaded        | NIL        | NIL     |
| statically linked | :STATIC    | NIL     |
| dlopen()'d        | :EXTERNAL  | non-NIL |

### DIFF
--- a/doc/cffi-manual.texinfo
+++ b/doc/cffi-manual.texinfo
@@ -6062,7 +6062,7 @@ CFFI> (let ((lib (load-foreign-library '(:framework "OpenGL"))))
 
 @Macro{define-foreign-library name-and-options @{ load-clause @}* @res{} name}
 
-name-and-options ::= name | (name &key convention search-path)
+name-and-options ::= name | (name &key canary convention search-path)
 load-clause ::= (feature library &key convention search-path)
 
 @subheading Arguments and Values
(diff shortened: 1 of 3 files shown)


doc/cffi-manual.texinfo
```diff
@@ -6077,6 +6077,12 @@ A feature expression.
 @item library
 A library designator.
 
+@item canary
+A string denoting a foreign symbol that will be searched in core
+before attempting to load the library. If that symbol is found, the
+library is assumed to be statically linked and
+@code{load-foreign-library} only marks the library as loaded.
```
luismbo, line 6084: Who would use this? Wrappers aren't defined via define-foreign-library.

sionescu: I compiled an SBCL that statically links libfixposix and openssl (https://cdn.cddr.org/ci/sbcl-2.0.4+r02-binary.tar.bz2), and I'd like to start using that: https://github.com/sionescu/iolib/commit/3bf5fc6883c8c2705744bf2d42522469c3f14cb7.

luismbo: It'd be nice to link somewhere (do SBCL or ECL or something document this?) or add a footnote about how one might statically link a library like this.

sionescu: I added a note to that effect, but I don't think we should add links because they can get stale and SBCL's is effectively undocumented even though supported.

sionescu: We could also use this for wrappers, but I don't think it's the right moment. We need a more comprehensive static linking story that supports lots of use cases, and that takes much work.

luismbo: Agreed.

PR #161, opened 2020-05-22
Do not ignore cflags of c-file.

ralt:
For example:

```lisp
(defsystem "foo"
  :components ((:c-file "foo" :cflags ("-lz" "-lm"))))
```

The slot is already there, but isn't actually used. It's useful when
custom compilation options need to be passed.

Additionally, the cflags slot is removed from o-file, as it does not
make sense there. Nor is it used, anyway.


luismbo:
A test case would be nice, if you're up for it. (Perhaps we can/should use this for libtest.c, for instance.) Thanks!

PR #162, opened 2020-05-23
Register the c-file libraries as :static.

ralt:
Note, however, that it is not running LOAD-FOREIGN-LIBRARY, as that
would have it reloaded when the image is restored. For example, SBCL
stores the loaded foreign libraries in its "shared objects" concept,
and tries to restore them all by default when restoring. See also
https://github.com/cffi/cffi/pull/157/ for a similar effect.

By registering the c-file libraries as :static, using the
recently-merged FOREIGN-LIBRARY-LOAD-STATE, we can make CFFI not use
dlopen() after the image is restored, when the library will be needed.

### DIFF
--- a/src/package.lisp
+++ b/src/package.lisp
@@ -121,11 +121,13 @@
    #:foreign-library-loaded-p
    #:list-foreign-libraries
    #:define-foreign-library
+   #:register-foreign-library
    #:load-foreign-library
    #:load-foreign-library-error
    #:use-foreign-library
    #:close-foreign-library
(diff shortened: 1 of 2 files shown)


sionescu:
This behavior is incorrect: the ASDF operation is generating a .so, which is a shared object not a static archive. You can't just declare it to be statically linked.

ralt:
Hm, indeed, it only works in the context of static-program-op. I'll try to think of a way to apply this state only for this operation.

PR #163, opened 2020-05-25
Always close and load foreign libraries when dump/restoring.

ralt:
Supersedes #162 and #157

### DIFF
--- a/grovel/grovel.lisp
+++ b/grovel/grovel.lisp
@@ -768,6 +768,7 @@ string."
                      (cffi:define-foreign-library
                          (,named-library-name
                           :type :grovel-wrapper
+                          :reload-on-restart nil
                           :search-path ,(directory-namestring lib-file))
                        (t ,(namestring (make-so-file-name lib-soname))))
                      (cffi:use-foreign-library ,named-library-name))
--- a/src/libraries.lisp
+++ b/src/libraries.lisp
@@ -231,7 +231,7 @@ all libraries are returned."
           spec))
 
 (defmethod initialize-instance :after
-    ((lib foreign-library) &key canary search-path
+    ((lib foreign-library) &key canary (reload-on-restart t) search-path
      (cconv :cdecl cconv-p)
      (calling-convention cconv calling-convention-p)
      (convention calling-convention))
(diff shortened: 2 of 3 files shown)


ralt:
1. ha! this is actually why I wanted to expose register-foreign-library. The name argument evaluation. Should I just add an argument to load-foreign-library?
2. How should we do that? A new special variable?
3. That makes sense.

luismbo:
1. Yes, I think so.
2. Maybe. If one can remove the our hook from the uiop:*image-restore-hook* list, maybe that's enough.

sionescu:
@luismbo load-foreign-library works in a DWIM mode:
 * if the first argument is a symbol, it assumes it names a library previously defined via register-foreign-library, then simply loads that using the options saved in the the library CLOS object
 * if it's a pathname designator, it works in "simple mode" and registers a library using default values, which it then loads

If load-foreign-library were to take a reload-on-restart argument, it would do a third thing: some of the load parameters would come from the library object named by the symbol, and some from the function's argument list. In case of discrepancy, which value takes priority ? Are discrepant values saved back into the library object or do they take effect transiently ?

This is all becoming too confusing and a potential source of bugs so I would opt for requiring all non-default options to be set via register-foreign-library.

sionescu:
Before digging ourselves deeper into ASDF, I would suggest taking a step back and thinking about the consequences of this integration with ASDF.
By mandating the use of `uiop:*image-restore-hook*` we prevent other image saving libraries from working properly without binding them to the ASDF internals.
Do we want to tell Buildapp users that they can't save an image that uses CFFI, because CFFI now mandates ASDF not just for loading code but for saving images too ?

luismbo:
I don't see how we'd be mandating ASDF. We're just implementing useful behaviour for people using ASDF to save images. We can implement or help implement similar behaviour for other libraries.

sionescu:
It mandates ASDF by using part of it (uiop:*image-restore-hook*).
I think CFFI should only implement the mechanism(s) that allow several image-building use cases and let the image-saving use CFFI APIs instead, so avoid pushing to uiop:*image-restore-hook* and not reference ASDF symbols as much as possible.

luismbo:
OK. Seems easy to slap a #+asdf there!

sionescu:
That won't work because I will still use ASDF to compile the code, just not to save the image.

luismbo:
If you're not using ASDF to save the image, then `uiop:*image-restore-hook*` won't affect you.

sionescu:
Ok, then just put #+asdf where needed.

sionescu:
Ok, then just put #+asdf where needed.

ruricolist:
What is the status of this PR? Is it just waiting on `#+asdf` feature expressions?

luismbo:
No, there a few open issues in the discussion. (Plus documentation.)

PR #164, opened 2020-05-28
Add a test making sure :cflags of :c-file is used.

ralt:
I'm not exactly sure where those tests are run, but I've been using this to confirm it works:

```
$ sbcl --eval '(ql:quickload :cffi-grovel)' --eval '(ql:quickload :cffi-tests/example)' --eval '(asdf:operate :static-program-op :cffi-tests/example)'
$ ~/.cache/common-lisp/sbcl-2.0.4-linux-x64/[...]/cffi/cffi-tests--example 
hello, world!
```

### DIFF
--- a/examples/test.c
+++ b/examples/test.c
@@ -1,0 +1,4 @@
+int mytest()
+{
+	return TEST;
+}
--- a/cffi-tests.asd
+++ b/cffi-tests.asd
@@ -86,6 +86,7 @@
      ((:file "package")
       (:cffi-wrapper-file "wrapper-example" :depends-on ("package"))
       (:cffi-grovel-file "grovel-example" :depends-on ("package"))
+      (:c-file "test" :cflags ("-DTEST=1"))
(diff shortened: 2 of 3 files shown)


ralt:
cc @luismbo given your previous request :) instead of trying to link libtest.c, I simply added a preprocessor variable, much simpler to test, and does the job just as well.

examples/test.c
```diff
@@ -0,0 +1,4 @@
+int mytest()
```
luismbo, line 1: mytest, you can find a better name than that :D maybe `test_c_file_cflags`?

examples/test.c
```diff
@@ -0,0 +1,4 @@
...
+{
+	return TEST;
```
luismbo, line 3: What happens if TEST is undefined? Maybe add a fallback #ifndef and call this `TEST_CFFI_CFLAG` or something.

PR #166, opened 2020-06-02
abcl: implement MAKE-SHAREABLE-BYTE-VECTOR

easye:
In abcl-1.7.0, CL:MAKE-ARRAY is able create and share ```malloc()```d memory
for java.nio.ByteVector allocated by JNA via use of the ```:NIO-BUFFER```
argument. Whether this feature is present in the current
implementation is conditionalized on the presence ```:NIO``` in
```CL:*FEATURES*```.

We retain the old behavior of copying memory for
WITH-POINTER-TO-VECTOR-DATA when passed a non-shareable byte vector as
a plausible fallback strategy.

supersedes #165

### DIFF
--- a/src/cffi-abcl.lisp
+++ b/src/cffi-abcl.lisp
@@ -69,8 +69,6 @@
    #:native-namestring
    #:%mem-ref
    #:%mem-set
-   ;; #:make-shareable-byte-vector
-   ;; #:with-pointer-to-vector-data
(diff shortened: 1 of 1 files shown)


luismbo:
Cool, thanks!

easye:
Binaries for ABCL with ```:nio``` now available as <https://abcl.org/releases/>.

PR #167, opened 2020-06-30
Solaris ld doesn't do --export-dynamic

jesseoff:
I believe all symbols are default exported w/the Solaris linker

### DIFF
--- a/toolchain/c-toolchain.lisp
+++ b/toolchain/c-toolchain.lisp
@@ -116,7 +116,7 @@
        (setf (symbol-value sym)
              (if normalizep (normalize-flags linkset val) val))))
     (setf *ld* *cc*
-          *ld-exe-flags* `(,@*cc-flags* #-(or sunos solaris darwin) "-Wl,--export-dynamic")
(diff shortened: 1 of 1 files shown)


luismbo:
Which Lisps use `:sunos` and which use the `:solaris` feature?

jesseoff:
I only have two successfully running on my Solaris machine.  sbcl uses :sunos and ccl uses :solaris

priyadarshan:
> 
> 
> Which Lisps use `:sunos` and which use the `:solaris` feature?

If I may butt in, on [LispWorks](http://www.lispworks.com/documentation/lw70/LW/html/lw-373.htm) is `:solaris2`

luismbo:
@jesseoff could you add this to trivial-features so that we can use `:solaris` everywhere?

jesseoff:
Will do.  Maybe I'll try to see if ecl also can build and what it uses.

jesseoff:
I have a pending pull request in trivial-features to normalize :sunos in *features* (the default for sbcl)  If that gets approved, I'll update this to use only :sunos

PR #169, opened 2020-08-02
Add arm64 as a platform (confirmed that this will enable at least min…

pbella:
…imally running cl-cffi-gtk on Debian on arm64)


luismbo:
Please don’t use cffi-features. It’s been deprecated in favor of trivial-features for 13 years or so.

PR #170, opened 2020-08-21
Add all types defined in stdint.h

Shinmera:
This (finally) adds standard types defined in `stdint.h` to CFFI, meaning projects no longer manually have to define things such as `size_t`. Tests are present to compare the types that do not have explicit bittage. The tests pass on my linux AMD64 machine, though it would also need to be run on other platforms to make sure that we're doing the right thing here.

Also, CFFI will now emit a compiler or style warning if one of the precise bittage types cannot be found, and thus cannot be defined.

### DIFF
--- a/src/types.lisp
+++ b/src/types.lisp
@@ -1034,24 +1034,36 @@ The buffer has dynamic extent and may be stack allocated."
 (defctype :llong  :long-long)
 (defctype :ullong :unsigned-long-long)
 
+(defmacro defctype-matching (name size-or-type base-types &key (match-by '=))
+  (let* ((target-size (typecase size-or-type
+                        (integer size-or-type)
+                        (t (foreign-type-size size-or-type))))
+         (matching-type (loop for type in base-types
+                              for size = (foreign-type-size type)
+                              when (funcall match-by target-size size)
+                              return type)))
+    (if matching-type
+        `(defctype ,name ,matching-type)
+        `(warn "Found no matching type of size ~d in~%  ~a"
+               ,target-size ',base-types))))
+
 ;;; We try to define the :[u]int{8,16,32,64} types by looking at
 ;;; the sizes of the built-in integer types and defining typedefs.
-(eval-when (:compile-toplevel :load-toplevel :execute)
-  (macrolet
-      ((match-types (sized-types mtypes)
-         `(progn
-            ,@(loop for (type . size-or-type) in sized-types
-                    for m = (car (member (if (keywordp size-or-type)
-                                             (foreign-type-size size-or-type)
-                                             size-or-type)
-                                         mtypes :key #'foreign-type-size))
-                    when m collect `(defctype ,type ,m)))))
-    ;; signed
-    (match-types ((:int8 . 1) (:int16 . 2) (:int32 . 4) (:int64 . 8)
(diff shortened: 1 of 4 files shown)


src/types.lisp
```diff
@@ -1034,24 +1034,64 @@ The buffer has dynamic extent and may be stack allocated."
 (defctype :llong  :long-long)
 (defctype :ullong :unsigned-long-long)
 
+(defmacro defctype-matching (name size-or-type base-types &key (match-by '=))
+  (let* ((target-size (typecase size-or-type
+                        (integer size-or-type)
+                        (T (foreign-type-size size-or-type))))
```
luismbo, line 1040:
```suggestion
                        (t (foreign-type-size size-or-type))))
```

src/types.lisp
```diff
@@ -1034,24 +1034,64 @@ The buffer has dynamic extent and may be stack allocated."
...
+         (matching-type (loop for type in base-types
+                              for size = (foreign-type-size type)
+                              do (when (funcall match-by target-size size)
+                                   (return type)))))
```
luismbo, line 1044:
```suggestion
                              when (funcall match-by target-size size)
                                return type)))
```

or maybe `(find target-size base-types :key #'foreign-type-size :test match-by)`?

src/types.lisp
```diff
@@ -1034,24 +1034,64 @@ The buffer has dynamic extent and may be stack allocated."
...
+    (if matching-type
+        `(defctype ,name ,matching-type)
+        `(warn #+sbcl 'sb-int:simple-compiler-note #-sbcl 'alexandria:simple-style-warning
```
luismbo, line 1047: sb-int is an internal package isn't it? What's the rationale anyway?

luismbo (changes requested):

src/types.lisp
Shinmera: A style warning doesn't seem quite appropriate, and neither does a full warning. The compiler note seems like the closest fitting condition type.

luismbo: A full warning would likely result in a bug report which would be nice. In any case, if we could avoid using SBCL internals, that'd be nice. Don't want to anger Douglas or Stas. :D

Shinmera:
Comments should be addressed now.

sionescu:
I don't think it's worth much defining the fast/least variants and (u)intmax types. They can't be used in APIs because they're implementation-dependent, rather than arch- or os-dependent, so we can ignore them for the purposes of a FFI library.

sionescu:
And the current definitions are incorrect because they must reflect the user's C compiler. For example, according to [this](https://stackoverflow.com/questions/20459513/is-intmax-t-the-same-as-long-long-int), a compiler is allowed to make uintmax_t 128bits.

Shinmera:
I've removed the types as requested, and have added the ``offset`` type, which is one I missed (and also one used by posix).

sionescu:
That looks good. Can you cleanup the test files as well and squash the commits into one ?

sionescu:
Thanks. Add tests for (u)intptr_t as well and we should be good to go.

Shinmera:
Here we go then!

sionescu:
Thanks.

PR #171, opened 2020-11-09
support arrays in structs

death:
Prior to this change, foreign-struct-slot-value of an
aggregate-struct-slot would return a pointer to the beginning of the
array, and (setf foreign-struct-slot-value) would result in an error.

With this change, foreign-struct-slot-value and (setf
foreign-struct-slot-value) take the aggregate slot's count into
account.  If the count is 1, the array element's value is returned.
Otherwise, a list of the array element values is returned.

This change is not backwards compatible because
foreign-struct-slot-value of an aggregate-struct-slot no longer
returns a pointer, but a value, as can be expected from its name, from
the generic function's docstring, and from behavior of the method
specialized on simple-struct-slot.  I had to change the union.1 test
since it expected the old behavior.  I think this change also
indicates that the new behavior makes sense.

### DIFF
--- a/src/types.lisp
+++ b/src/types.lisp
@@ -569,18 +569,25 @@ The foreign array must be freed with foreign-array-free."
   ((count :initarg :count :accessor slot-count))
   (:documentation "Aggregate structure slots."))
 
-;;; Since MEM-REF returns a pointer for struct types we are able to
-;;; chain together slot names when accessing slot values in nested
-;;; structures.
 (defmethod foreign-struct-slot-value (ptr (slot aggregate-struct-slot))
-  "Return a pointer to SLOT relative to PTR."
-  (convert-from-foreign (inc-pointer ptr (slot-offset slot))
-                        (slot-type slot)))
+  "Return the value of an aggregate SLOT from a struct at PTR."
+  (let ((type (slot-type slot))
+        (input-ptr (inc-pointer ptr (slot-offset slot))))
+    (if (= 1 (slot-count slot))
+        (convert-from-foreign input-ptr type)
+        (loop for i below (slot-count slot)
+              collect (mem-aref input-ptr type i)))))
 
 (defmethod foreign-struct-slot-value-form (ptr (slot aggregate-struct-slot))
   "Return a form to get the value of SLOT relative to PTR."
-  `(convert-from-foreign (inc-pointer ,ptr ,(slot-offset slot))
-                         ',(slot-type slot)))
+  (let ((type (slot-type slot))
+        (input-ptr `(inc-pointer ,ptr ,(slot-offset slot)))
+        (count (slot-count slot))
+        (i (gensym)))
+    (if (= 1 count)
+        `(convert-from-foreign ,input-ptr ',type)
+        `(loop for ,i below ,count
+               collect (mem-aref ,input-ptr ',type ,i)))))
 
 (defmethod translate-aggregate-to-foreign (ptr value (type foreign-struct-type))
   ;;; FIXME: use the block memory interface instead.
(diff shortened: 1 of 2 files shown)


src/types.lisp
```diff
@@ -569,18 +569,25 @@ The foreign array must be freed with foreign-array-free."
...
-  "Return a pointer to SLOT relative to PTR."
-  (convert-from-foreign (inc-pointer ptr (slot-offset slot))
-                        (slot-type slot)))
+  "Return the value of an aggregate SLOT from a struct at PTR."
+  (let ((type (slot-type slot))
+        (input-ptr (inc-pointer ptr (slot-offset slot))))
+    (if (= 1 (slot-count slot))
+        (convert-from-foreign input-ptr type)
+        (loop for i below (slot-count slot)
+              collect (mem-aref input-ptr type i)))))
```
luismbo, line 579: I wonder if we should instead deprecate the `:count` option and convert `(x foo :count n)` slots to `(x (:array foo n))`. That way we wouldn't need specific code for handling `:count`.

Did you know you can do this?

```lisp
CFFI> (defcstruct test
        (x (:array :int 2)))
(:STRUCT TEST)
CFFI> (with-foreign-object (x '(:struct test))
        (setf (foreign-slot-value x '(:struct test) 'x) #(42 42))
        (foreign-slot-value x '(:struct test) 'x))
#(42 42)
```

(I didn't remember it was possible when I started reviewing your PR.)

src/types.lisp
```diff
@@ -590,15 +597,28 @@ The foreign array must be freed with foreign-array-free."
...
-                                  value
-                                  (parse-type (slot-type slot))))
+  (let ((type (slot-type slot))
+        (output-ptr (inc-pointer ptr (slot-offset slot))))
+    (if (= 1 (slot-count slot))
+        (translate-aggregate-to-foreign output-ptr value (parse-type type))
+        (loop for i below (slot-count slot)
+              for v in value
+              do (setf (mem-aref output-ptr type i)
+                       (translate-to-foreign v type))))))
```
luismbo, line 607: I think this should use `convert-into-foreign-memory` instead, but see my previous about piggybacking on the `:array` type.

luismbo (changes requested):
I'm generally weary of breaking backwards compatibility at this point in CFFI's life, but this generally makes sense and we've done something similar for nested struct slots in the past, so this would make things more consistent.

I wonder whether we should deprecate the `:count` option and instead review and document the `:array` type.

Should we go ahead with making `:count` saner, some tests would be nice.

src/types.lisp
death: I didn't know about `:array`, but I'm not sure it works with C functions that take `(:struct test)`. If a C function returns a `(:struct test)` I will get, say, a list `(X #(1 2))`. If I then pass it to a C function, CFFI will convert the array to a foreign value and get a pointer (which is not the right thing to do), and then it will try to set the foreign slot's value to this pointer, but the slot is an `aggregate-struct-slot` so the value expected is a Lisp array.

I wondered how much code uses `:count` vs. `:array`. I scanned some of the Lisp libraries I have locally and came up with [this](https://gist.github.com/death/a7403e30d0b6eb3b759ad5901e745a0c). To generate this report I wrote [formgrep](https://github.com/death/formgrep). It's not surprising that `:array` is not used much, given that it's not documented, but maybe this should help a decision regarding deprecation of `:count`.

luismbo: The issue with `:array` you describe should be fixed as follows:

```lisp
(defmethod translate-into-foreign-memory ((object list)
                                          (type foreign-struct-type)
                                          p)
  (unless (bare-struct-type-p type)
    (loop for (name value) on object by #'cddr
          do (setf (foreign-struct-slot-value p (get-slot-info type name))
                   value))))
```

After that fix, the following snippet does what we expect, right?

```
CFFI> (defcstruct foo (x (:array :int 2)))
(:STRUCT FOO)
CFFI> (with-foreign-object (x '(:struct foo))
        (setf (mem-ref x '(:struct foo))
              (list 'x #(21 42)))
        (list (mem-aref x :int 0) (mem-aref x :int 1)))
(21 42)
```

Regarding :count vs :array, yeah, that's not unexpected. Should we go the deprecation route, we should deprecate in the documentation first, after a few years add a deprecating style warning, then a warning, then an error, then remove it. It shouldn't take more than 20 years. :D

death: It seems there are still unexpected values; it probably needs some tests with actual C functions taking and returning such structures.

PR #172, opened 2020-11-09
fix :string return type in cffi-libffi

death:
Prior to this change, foreign-string-to-lisp would receive a pointer
to the string pointer, and not the string pointer itself.

This issue may affect other types as well, but I did not inquire
further.

### DIFF
--- a/libffi/funcall.lisp
+++ b/libffi/funcall.lisp
@@ -70,7 +70,7 @@
    types
    return-type
    (if (or (eql return-type :void)
-           (typep (parse-type return-type) 'translatable-foreign-type))
+           (typep (parse-type return-type) 'foreign-struct-type))
        call-form
        ;; built-in types won't be translated by
        ;; expand-from-foreign, we have to do it here


libffi/funcall.lisp
```diff
@@ -70,7 +70,7 @@
    types
    return-type
    (if (or (eql return-type :void)
-           (typep (parse-type return-type) 'translatable-foreign-type))
+           (typep (parse-type return-type) 'foreign-struct-type))
```
attila-lendvai, line 73: i have lost context long ago, but i had a gut feeling looking at this: maybe the way to address this is that foreign-struct-type should inherit from translatable-foreign-type ?

or if that is not an option, then maybe check for both?

it would be nice to have a few tests that cover at least the basics of this...

attila-lendvai: err, scratch the test part: there is tests/fsbv.lisp.

did you run the tests with this change? could you please add a test that fails prior to the change?

death: In libffi calls, we allocate a foreign object and pass a pointer to it to libffi so that it can fill the object with the return value.

After the call we need to dereference that pointer to get at the foreign object. We need to do that for translatable types like `:string` (where the foreign object is a char pointer) or `:enum` too.

We don't need to dereference the pointer for foreign structs (or unions) because its `translate-from-foreign` method expects a pointer.

I added a test to demonstrate the issue with the `:string` return type.

luismbo: So, maybe the comment

```lisp
       ;; built-in types won't be translated by
       ;; expand-from-foreign, we have to do it here
```

needs to be updated. What's going on here is simply that non-struct types have an extra indirection.

This code is a bit confusing because it's chaining `expand-from-foreign` (via `translate-objects`) and `mem-ref` together. Perhaps reusing `translate-objects` here is doing more harm than good. How about something along these lines?

```lisp
(defun translate-objects-ret (symbols function-arguments types return-type call-form)
  (cond (function-arguments
         (expand-to-foreign-dyn-indirect
          (car function-arguments)
          (car symbols)
          (list (translate-objects-ret (cdr symbols) (cdr function-arguments)
                                       (cdr types) return-type call-form))
          (parse-type (car types))))
        ((eq return-type :void)
         call-form)
        (t
         `(mem-ref ,call-form ,return-type))))
```

luismbo (changes requested):

PR #174, opened 2020-12-30
sbcl/call-within-initial-thread: don't interrupt random threads.

stassats:
Use sb-thread::*initial-thread*

For https://bugs.launchpad.net/cffi/+bug/1906993


sionescu:
Thanks, merged.

PR #175, opened 2020-12-31
Split *type-parsers* into three EQL hash tables

pfdietz:
Also add specialized lookup for the default table case.
This is noticeably faster in SBCL than the previous EQUAL hash table scheme.


sionescu:
Merged. Thanks.

PR #176, opened 2021-01-31
Fix macro discard-docstring

sdilts:
When a list has one element, `(cdr list)` is nil, so it  must be negated in `discard-docstring` for the condition to ever be true in the correct circumstances.

Without this change, calls to `defcstruct` and `defcunion` with a docstring and no slots fails with the same error:
``` common lisp
(cffi:defcstruct foo "docstring")

;; Error while parsing arguments to DESTRUCTURING-BIND:
;; invalid number of elements in
;;   "docstring"
;; to satisfy lambda list
;;   (CFFI::SLOTNAME TYPE &KEY (COUNT) CFFI::OFFSET):
;; at least 2 expected, but got a non-list
;;  [Condition of type SB-KERNEL::ARG-COUNT-ERROR]

(cffi:defcunion bar "Union docstring")

;; Error while parsing arguments to DESTRUCTURING-BIND:
;; invalid number of elements in
;;   "Union docstring"
;; to satisfy lambda list
;;   (CFFI::SLOTNAME TYPE &KEY (COUNT)):
;; at least 2 expected, but got a non-list
;;  [Condition of type SB-KERNEL::ARG-COUNT-ERROR]
```

I'm not sure how to write a test for a macro, so I didn't add one.

### DIFF
--- a/src/utils.lisp
+++ b/src/utils.lisp
@@ -27,10 +27,10 @@
 
 (in-package #:cffi)
 
-(defmacro discard-docstring (body-var &optional force)
+(defmacro discard-docstring (body-var)
   "Discards the first element of the list in body-var if it's a
-string and the only element (or if FORCE is T)."
-  `(when (and (stringp (car ,body-var)) (or ,force (cdr ,body-var)))
+string."
+  `(when (stringp (car ,body-var))
      (pop ,body-var)))
 
 (defun single-bit-p (integer)


sdilts:
Interestingly, the commit (ffd4f89783e7dc01ff72420696b935cab9614653) adding this check seems to have this bug too, muddying the original intention. Either that, or the docstring is just wrong. The tests that were added in that same commit pass still. so something else might have changed to stop the original bug from occurring. 

I haven't tried it, but with that discovery, could we remove the `force` param and have everything still work?

luismbo:
https://github.com/cffi/cffi/commit/ffd4f89783e7dc01ff72420696b935cab9614653 alludes to `defcallback` and that's where that behaviour of not discarding the docstring if the body contained nothing else was relevant. `defcallback` was since changed to use `alexandria:parse-body`, so I think your idea of removing the `force` parameter and the list check is the way to go.

Out of curiosity, are you actually using an empty structure?

sdilts:
> Out of curiosity, are you actually using an empty structure?

I'm using them for opaque structures, yes. The library I'm wrapping has code like this:
``` c
/**
 * @struct xkb_keymap
 * Opaque compiled keymap object.
 *
 * The keymap object holds all of the static keyboard information obtained
 * from compiling XKB files.
 *
 * A keymap is immutable after it is created (besides reference counts, etc.);
 * if you need to change it, you must create a new one.
 */
struct xkb_keymap;
```
And the tool I'm using to create the bindings generates this code: 
``` lisp
(cffi:defcstruct xkb-keymap
  "@struct xkb_keymap
Opaque compiled keymap object.

The keymap object holds all of the static keyboard information obtained
from compiling XKB files.

A keymap is immutable after it is created (besides reference counts, etc.);
if you need to change it, you must create a new one.")
```
So without this change, the generated bindings don't work. Having empty structs is invalid C, but C code I'm translating from is definitely valid. Is this a valid way to do it CFFI-wise?

sionescu:
IMO the best solution is to use the parse-declarations library to properly detect docstrings.

luismbo:
@sionescu that would yield the wrong results since `defcstruct` semantics and `defun` are not the same, which is what @sdilts is trying to fix here.

luismbo:
@sdilts an empty structure might not be standard C, but that library is using it and GCC accepts it. I think CFFI treats it the same as GCC: the structure has a size of 0. I suppose that defcstruct lets you use the `(:pointer (:struct xkb-keymap))` type, which is currently identical to `:pointer` but, of course, clearer.

PR #177, opened 2021-02-26
on arm64 macos check /opt/homebrew/lib for libraries first

slyrus:
* This is the (current) default path for arm64 homebrew
   libraries. Try this before the rest.


luismbo:
Thanks. It'd be nice to check that all supported Lisps do push `:arm64` to `*features*` and amend `trivial-features` (or simply open an issue) if needed!

PR #178, opened 2021-03-24
Grovel: Detect :float and :double in the :auto type

leosongwei:
extend the existing C semantics trick to detect float types in the grovel.


luismbo:
@leosongwei 谢谢！

luismbo:
@leosongwei had to revert this due to https://bugs.launchpad.net/cffi/+bug/1922428

PR #180, opened 2021-04-06
add prologue callback right next to epilogue callback

some-mthfka:
I need this for https://github.com/some-mthfka/hu.dwim.sdl, where one package (aka sdl-ttf) relies on the definitions from another package (sdl-core), and I import all the definitions in prologue. I can't do it in prelude because the generated file redefines the package that it generates.

Or maybe there's indeed another way of doing it?


luismbo:
You’ll have to explain your use case in a bit more. I didn’t understand your description. Perhaps @attila-lendvai can chime in too.

some-mthfka:
@luismbo So, what happens is: the main sdl lib defines certains types/structure such as SDL_Color. Now, SDL-TTF is generated in a seperate package, but some of its function such as render-text-solid rely on these definitions and if it doesn't have them, there will be comments in the generated code that sdl-color was not found and ttf-render-text-solid was skipped. And so I import all the symbols from the core package into the packages that depend on it (well, ommiting the clashing symbols like init or quit) and that solves the problem.

src/c2ffi/generator.lisp
```diff
@@ -537,7 +537,7 @@ target package."
   (with-standard-io-syntax
     (with-input-from-file (in c2ffi-spec-file :external-format (uiop:encoding-external-format :utf-8))
       (with-output-to-file (*c2ffi-output-stream* output :if-exists :supersede
-                            :external-format (uiop:encoding-external-format output-encoding))
+                                                         :external-format (uiop:encoding-external-format output-encoding))
```
attila-lendvai, line 540: for `git blame` it's better to avoid recording changes to parts that you don't need to touch.

some-mthfka: Would it solve the problem if I change the indentation back to how it was and push again? Or would it be easier to make another pull request at this point?

some-mthfka: @attila-lendvai indentation restored, seems to have done the trick

attila-lendvai: thanks!

FYI, you can `git push -f` into branches on github, and the relevant PR gets updated automatically.

some-mthfka: @attila-lendvai my bad, got it now!

attila-lendvai:
FTR, this PR is needed for this issue: https://github.com/hu-dwim/hu.dwim.sdl/issues/2

but for now i'm reluctant to proceed on this. further discussion is under that issue.

PR #181, opened 2021-04-17
Fix translation of struct-typedefs (take 2)

luismbo:
CFFI now always translates aggregate struct values to lisp property
lists.  Earlier struct values which were defined by types introduced
by defctype were translated as foreign pointers.

One consequence of this change is that defcvar to a foreign struct
will always produce an immutable lisp list.  To modify the struct the
user first has to get at the pointer to the struct via
FOREIGN-SYMBOL-VALUE or GET-VAR-POINTER.  To modify nested structs the
user has to get at the pointer to the field via FORIEGN-SLOT-POINTER,
- or use the the (:POINTER VAR) syntax of WITH-FOREIGN-SLOTS.

* src/early-types.lisp: (defctype): use translatable-foreign-type
instead of enhanced-foreign-type.

* tests/structs.lisp: New tests: (misnamed) STRUCT-VALUES.FSBV.1
STRUCT-VALUES.FSBV.1 Updated tests: STRUCT.ALIGNMENT.1
STRUCT.ALIGNMENT.2 STRUCT.ALIGNMENT.3 STRUCT.ALIGNMENT.4
STRUCT.ALIGNMENT.5 STRUCT.ALIGNMENT.6 STRUCT.ALIGNMENT.7
STRUCT.NESTED-SETF STRUCT.ALIGNMENT.8

### DIFF
--- a/libffi/funcall.lisp
+++ b/libffi/funcall.lisp
@@ -70,7 +70,8 @@
    types
    return-type
    (if (or (eql return-type :void)
-           (typep (parse-type return-type) 'translatable-foreign-type))
+           (typep (parse-type (canonicalize-foreign-type return-type))
+                  'translatable-foreign-type))
        call-form
        ;; built-in types won't be translated by
        ;; expand-from-foreign, we have to do it here
--- a/src/early-types.lisp
+++ b/src/early-types.lisp
@@ -705,7 +705,7 @@ Signals an error if the type cannot be resolved."
   (declare (ignore documentation))
   (warn-if-kw-or-belongs-to-cl name)
   (let* ((btype (parse-type base-type))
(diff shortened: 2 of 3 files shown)


luismbo:
TODO: test the failures identified in #152 (and possibly add more tests)

enometh:
[ sorry I thought I was commenting on #181 after you closed #182 but somehow the web ui decided to create a new request #183 which I closed.

As I said on #182 - this commit should be split into two parts - one to fix the decades old bug in ffi
and the second to revert the reverted commit.

merging the two into a single commit unnecessarily obscures the issue
I don't believe they should be merged as it be outright misleading
and the issues are orthogonal

PR #182, opened 2021-04-18
Pull 181 tmp

enometh:
Identical in content with pull-181 but splits the fix to
libffi (translate-objects-ret):  from the commit
reverted in
https://github.com/cffi/cffi/pull/152#issuecomment-821581524

### DIFF
--- a/libffi/funcall.lisp
+++ b/libffi/funcall.lisp
@@ -70,7 +70,8 @@
    types
    return-type
    (if (or (eql return-type :void)
-           (typep (parse-type return-type) 'translatable-foreign-type))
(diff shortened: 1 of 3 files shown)


luismbo:
I think I'd rather have them in a single commit since they're not quite independent.

enometh:
Why? it would only hide the (potential) original bug in CFFI -  the commit itself as in the earlier pull request should not fix an independent original bug as it is completely unrelated.

I believe Steve Nunez in CFFI-DEVEL hit the original bug - which was again unrelated to this reverted commit.

I would prefer that the actual issues not be deliberately obscured by merging the commits

PR #185, opened 2021-05-15
Signal an error when a foreign function is not found on ABCL.

jathd:
When a foreign function is not found by `find-foreign-function` (either because
the function doesn't exist, or because the library is not loaded), `nil` is
returned, so `%%foreign-funcall` is called with `nil` as a function, which
ultimately leads to a null-pointer exception.

Better to signal an error on the lisp side.


luismbo:
Thanks @jathd

luismbo:
cc @easye

PR #187, opened 2021-09-06
src/types.lisp: (with-foreign-slots) enhance

enometh:
Regularise syntax on the binding forms of WITH-FOREIGN-SLOTS to
conform to (name value)

(with-foreign-slots (vars ptr type) body) - Now each var can take one
of these forms:

SLOT-NAME -- binds SLOT-NAME to (FOREIGN-SLOT-VALUE SLOT-NAME)

(:POINTER SLOT-NAME) -- binds SLOT-NAME to (FOREIGN-SLOT-POINTER
SLOT-NAME)

(VAR-NAME SLOT-NAME) -- binds VAR-NAME to (FOREIGN-SLOT-VALUE
SLOT-NAME)

(:POINTER (VAR-NAME SLOT-NAME)) -- binds VAR-NAME
to (FOREIGN-SLOT-POINTER SLOT-NAME)

(VAR-NAME (:POINTER SLOT-NAME)) -- binds VAR-NAME to
(FOREIGN-SLOT-POINTER SLOT-NAME)

### DIFF
--- a/src/types.lisp
+++ b/src/types.lisp
@@ -845,24 +845,37 @@ The foreign array must be freed with foreign-array-free."
 (defmacro with-foreign-slots ((vars ptr type) &body body)
   "Create local symbol macros for each var in VARS to reference
 foreign slots in PTR of TYPE. Similar to WITH-SLOTS.
-Each var can be of the form: 
-  name                       name bound to slot of same name              
-  (:pointer name)            name bound to pointer to slot of same name
-  (name slot-name)           name bound to slot-name
-  (name :pointer slot-name)  name bound to pointer to slot-name"
+Each var can be of the form:
+
+SLOT-NAME - SLOT-NAME is bound to the slot of the same
+name. i.e. to (FOREIGN-SLOT-VALUE SLOT-NAME)
+
+(:POINTER SLOT-NAME) - SLOT is bound to the pointer to the slot of the
+same name. i.e. to (FOREIGN-SLOT-POINTER SLOT-NAME)
+
+(VAR-NAME SLOT-NAME) - VAR-NAME is bound to (FOREIGN-SLOT-VALUE SLOT-NAME)
+
+(VAR-NAME (SLOT-NAME :POINTER)) - VAR-NAME is bound
+to (FOREIGN-SLOT-POINTER SLOT-NAME)
+"
+  (when (consp type)
+    (assert (not (eql (car type) 'quote))))
   (let ((ptr-var (gensym "PTR")))
     `(let ((,ptr-var ,ptr))
        (symbol-macrolet
            ,(loop :for var :in vars
-               :collect
-                 (if (listp var)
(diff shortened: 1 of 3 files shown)


src/types.lisp
```diff
@@ -845,24 +845,38 @@ The foreign array must be freed with foreign-array-free."
...
+                  `(,var (foreign-slot-value ,ptr-var ',type ',var))
+                  (if (eq (first var) :pointer)
+                      (if (atom (second var))
+                          `(,(second var) (foreign-slot-pointer ,ptr-var ',type ',(second var)))
+                          `(,(first (second var)) (foreign-slot-pointer ,ptr-var ',type ',(second (second var)))))
+                      (if (atom (second var))
+                          `(,(first var) (foreign-slot-value ,ptr-var ',type ',(second var)))
+                          (if (eq (first (second var)) :pointer)
+                              `(,(first var) (foreign-slot-pointer ,ptr-var ',type ',(second (second var))))
+                              (error  "Malformed slot specificcation ~a: must be VAR or (VAR NAME) or (:POINTER VAR) or (VAR (:POINTER NAME)) or (:POINTER (VAR NAME))" var))))))
```
luismbo, line 879: Specification is misspelled.

src/types.lisp
```diff
@@ -845,24 +845,38 @@ The foreign array must be freed with foreign-array-free."
...
+
+SLOT - SLOT is bound to the slot of the same
+name. i.e. to (FOREIGN-SLOT-VALUE SLOT)
+
+(:pointer SLOT) - SLOT is bound to the pointer to the slot of the same
+name. i.e. to (FOREIGN-SLOT-POINTER SLOT)
+
+(NAME SLOT) - NAME is bound to (FOREIGN-SLOT-VALUE SLOT)
+
+(:POINTER (NAME SLOT)) - NAME is bound to (FOREIGN-SLOT-POINTER SLOT)
```
luismbo, line 858: Maybe we can drop this one?

luismbo (changes requested):

src/types.lisp
luismbo: I.e. support only `(name (:pointer slot))`.

enometh: we support both (:pointer slot-name) as well as (slot-name :pointer) 
by means of the :pointer extension. Therefore it seemed like a good idea at that
time to include all these other forms for symmetry and completeness

I'm not sure if we want to support only (slot-name :poiinter) which would be only be
consistent.

Perhaps the docstring could refer to a subset of the supported forms to keep it simple
and the manual could explain the rationale and the remaining forms.

luismbo: I think supporting both was a mistake and neither is documented, so perhaps we can get away with removing `(slot-name :pointer)`.

enometh: I think it was in the manual - and to be  probably more consistent it would be better to remove (:pointer slot-name) - and retain (slot-name :pointer) as the LHS is the variable, wouldn't it?

PR #188, opened 2021-09-22
grovel: Error message for IN-PACKAGE with bad package designator

lukego:
Report a bad package designator in (IN-PACKAGE :FOO) as

    The name :FOO does not designate any package.

instead of

    The value
      NIL
    is not of type
      PACKAGE


luismbo:
Thanks!

PR #189, opened 2021-10-12
Fixed link-executable, added new syntax: pkg-config-libs, ld-exe-flags

ghollisjr:
After updating GCC (11.1.0) and HDF5 (1.12.1), the groveler no longer called the linker with arguments in the right order (--as-needed means elements from libraries that come before object files in the command string will be excluded, leading to "undefined reference to..." errors), and additionally there was a need to add linker flags for generating groveler executables, so I changed the following:

* Placed ld-exe-flags at the end of the link-executable command to ensure libraries come after objects
* Added new syntax pkg-config-libs with similar usage as pkg-config-cflags but for "pkg-config X --libs"
* Added new syntax ld-exe-flags with similar usage as cc-flags to explicitly add library flags


luismbo:
Can you drop the libffi commit? Also, are these compatible with older GCC versions?

ghollisjr:
Sure I'll create a new pull request for just the pkg-config & GCC argument order updates.

It should work with older GCC too, since the issue is that newer GCC has more picky behavior when the --as-needed argument is used as far as I'm aware.

PR #191, opened 2021-10-26
Fixed for new libffi version 8

ghollisjr:
Updating to the newest version of libffi prevents CFFI's libffi system from loading due to no longer finding the hard-coded versions of the libffi library, so this patch fixes that issue.

It might be better to solve this by just supporting both a fixed e.g. "/usr/lib/libffi.so" file as well as the list of specific versions, but this patch just adds the version 8 library paths.


attila-lendvai:
there was another PR, that one also updated :windows

closing this one, but thanks!

PR #192, opened 2021-11-19
Support libffi.so.8/libffi32.so.8 (unix) and libffi-8.dll (windows)

lukego:
I was tempted to load libffi.so and libffi32.so, to make this future-proof, but I didn't dare to be so bold :-)

~~Sorry, this is not tested yet, but I will do that if we don't have CI :-)~~ Seems to work!


lukego:
Yes, this works for me with `libffi.so.8` from libffi-3.4.2 on Linux/NixOS. I haven't tested `libffi32.so.8` or `libffi-8.dll`.

PR #193, opened 2021-11-24
Verify library path before loading

aniketnarvekar:


aniketnarvekar:
loading libcrypto in an unsafe way
This fixes the issue of CL+SSL library SBCL crash

avodonosov:
In https://github.com/cl-plus-ssl/cl-plus-ssl/issues/114#issuecomment-977925342 @daewok gives good reasons why this pull request is not suitable. Can be closed I believe.

It's interesting to note about the FIXME comment above the `load-foreign-library-path`:

```
;;; FIXME: haven't double checked whether all Lisps signal a
;;; SIMPLE-ERROR on %load-foreign-library failure.  In any case they
;;; should be throwing a more specific error.
```
As we learned in https://github.com/cl-plus-ssl/cl-plus-ssl/issues/138, the failure loading foreign lib can manifest itself as a process crash, and not as an error signaled in Lisp.

sionescu:
@avodonosov Sure, but that depends on what the libc does. The comment was about how the implementations translate an error returned by dlopen().

PR #194, opened 2021-12-01
Also scan LIBRARY_PATH when looking for shared objects.

attila-lendvai:
I didn't push it straight to master, maybe there's some issue with this that i'm not aware of...

see diff for the rationale.

@luismbo ?

### DIFF
--- a/src/libraries.lisp
+++ b/src/libraries.lisp
@@ -60,12 +60,17 @@
             #p"/usr/lib/")))
 
 (defvar *foreign-library-directories*
-  (if (featurep :darwin)
-      '((explode-path-environment-variable "LD_LIBRARY_PATH")
-        (explode-path-environment-variable "DYLD_LIBRARY_PATH")
-        (uiop:getcwd)
-        (darwin-fallback-library-path))
-      '())
+  ;; For LIBRARY_PATH see:
+  ;; https://gcc.gnu.org/onlinedocs/gcc/Environment-Variables.html#Environment-Variables
+  ;; The actual use-case is Guix, where `guix shell --development foo` updates this
+  ;; variable with a pointer to the profile's lib/ directory that contains the .so's.
+  (append
+   '((explode-path-environment-variable "LIBRARY_PATH"))
+   (when (featurep :darwin)
+     '((explode-path-environment-variable "LD_LIBRARY_PATH")
+       (explode-path-environment-variable "DYLD_LIBRARY_PATH")
+       (uiop:getcwd)
+       (darwin-fallback-library-path))))
   "List onto which user-defined library paths can be pushed.")
 
 (defun fallback-darwin-framework-directories ()


luismbo:
`dlopen()` should take `LIBRARY_PATH` into account on its own, shouldn’t it? In fact `* foreign-library-directories*` has been misimplemented from itches start as it doesn’t communicate those directories to `dlopen()` therefore it fails to find library dependencies when they exist.

sionescu:
`LIBRARY_PATH` is for the cc/ld command line tools. `dlopen()` only considers `LD_LIBRARY_PATH`. In trying to keep with the distinction between Ahead-of-time compilation and runtime linking, and to avoid potentially weird corner cases I don't think it's a good idea to search `LIBRARY_PATH` too.

luismbo:
Ah, missed that distinction. So, doesn’t guix’s environment affect dlopen()?

attila-lendvai:
> Ah, missed that distinction. So, doesn’t guix’s environment affect dlopen()?

in this regard Guix is like NixOS: any `.so` references are hardcoded into binaries with their full path to the immutable store location (e.g. along the lines of `/gnu/store/z4li262il798hbl0l1h1k3a5g7r6bffa-glibc-2.31/lib/libc-2.31.so`).

it's quite some fun compared to it still being a linux: you construct your system's definition in scheme (Guile) in a functional way, and then it "gets applied to the disc". if something goes wrong, then you can roll back to, or boot a previous profile. the package repository (together with the codebase of guix itself) is in a git repo. you have a time machine, and reproducible builds, etc.

`guix shell` only sets `LIBRARY_PATH`. in general guix doesn't mess with `LD_LIBRARY_PATH` whenever avoidable.

i briefly looked into setting `LD_LIBRARY_PATH` from `LIBRARY_PATH` at build time (and later restoring it at image start time), but for some reason CFFI fails to load `libffi.so` at build time, even though the file with the right name is there where `LD_LIBRARY_PATH` points to (as verified by `uiop:getenv` in the debugger context). maybe there's some Guix peculiarity behind this. i'll investigate further as time allows.

> That would work quite well for a C/C++ project and it would break if CFFI starts searching in LIBRARY_PATH.

thank you @sionescu ! that's something like what i was suspicious about, but i'm not well versed enough in the C toolchain to identify/predict such things.

attila-lendvai:
i think i have identified the crux of this issue:

> If, at the time that the program was started, the environment variable LD_LIBRARY_PATH was defined to contain [...]

it means that it's too late to set `LD_LIBRARY_PATH` from an image start hook, which is what i was trying in the build script.

unfortunately,

> The function dlopen() loads the dynamic library file named by the null-terminated string filename and returns an opaque "handle" for the dynamic library.

and therefore SBCL cannot (?) save the absolute path of the `.so`, and use that in `reopen-shared-objects` when the image is loaded. maybe it's not even desirable by default, because it would make images that much less portable.

so, the only, rather convoluted solution i can see is to also emit an `[image-name].library-path` file, and set `LD_LIBRARY_PATH` from the elisp code that starts sbcl with my image.

FTR, this is the starting version of the build script in question: https://hub.darcs.net/hu.dwim/hu.dwim.environment/browse/bin/build-development-image.sh

any hints for a simpler solution would be appreciated!

attila-lendvai:
FTR, this is my fix that seems to work: https://hub.darcs.net/hu.dwim/hu.dwim.environment/patch/87621c13906f1eddcd6a235dcc196d2306144b22

more complexity than i'd be happy about, but... it works.

PR #305, opened 2022-01-16
Handle numbers correctly in CFFI-GROVEL::C-WRITE

MrLightningBolt:
This allows the groveler to correctly handle cstruct definitions that include arrays


luismbo:
Can you add a unit test? That’d be nice.

MrLightningBolt:
Will this do? This is a basic test that can be extended to do more type groveling validation.

luismbo:
@MrLightningBolt thanks!

PR #308, opened 2022-01-31
Support variadic functions on SBCL-Darwin-arm64.

stassats:
Where variadic arguments are not just passed normally.

To pass all the tests the latest revision of SBCL from git is required, as the CFFI test suite uncovered a bug there.


luismbo:
Thanks. 👍

PR #310, opened 2022-02-07
Fixed ld-exe-flags for new GCC, added 'pkg-config --libs' support

ghollisjr:
After updating GCC (11.1.0) and HDF5 (1.12.1), the groveler no longer called the linker with arguments in the right order (--as-needed means elements from libraries that come before object files in the command string will be excluded, leading to "undefined reference to..." errors), and additionally there was a need to add linker flags for generating groveler executables, so I changed the following:

- Placed ld-exe-flags at the end of the link-executable command to ensure libraries come after objects
- Added new syntax pkg-config-libs with similar usage as pkg-config-cflags but for "pkg-config X --libs"
- Added new syntax ld-exe-flags with similar usage as cc-flags to explicitly add library flags

This should work for old and new GCC since as far as I'm aware it's a problem with newer GCC being more picky with the --as-needed flag, but I don't currently have an older GCC installed to explicitly test it.

### DIFF
--- a/grovel/grovel.lisp
+++ b/grovel/grovel.lisp
@@ -301,6 +301,29 @@ int main(int argc, char**argv) {
                 (t
                  (grovel-error "~a" message))))))))
 
+(define-grovel-syntax ld-exe-flags (&rest flags)
+  (appendf *ld-exe-flags* (parse-command-flags-list flags)))
+
+(define-grovel-syntax pkg-config-libs (pkg &key optional)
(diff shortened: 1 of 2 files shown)


ghollisjr:
UPDATE: I just tested this on GCC 8.3.0 and it works fine

ghollisjr:
UPDATE: Tested this again with GCC 12.1.0 and these features are still necessary to ensure CFFI works with both old and new GCC versions.

PR #312, opened 2022-02-09
store pointer to a foreign-alloc so expanded code can modify pointer

AeroNotix:
Without this it is possible that expanded code in `,@body` will perform pointer maths (e.g. `cffi:incf-pointer`) which causes a segv when freeing the original pointer, since it may no longer reference valid memory.

### DIFF
--- a/src/cffi-sbcl.lisp
+++ b/src/cffi-sbcl.lisp
@@ -155,11 +155,13 @@ SIZE-VAR is supplied, it will be bound to SIZE during BODY."
                  (,var (alien-sap ,alien-var)))
              (declare (ignorable ,size-var))
              ,@body)))
-      `(let* ((,size-var ,size)
-              (,var (%foreign-alloc ,size-var)))
-         (unwind-protect
-              (progn ,@body)
-           (foreign-free ,var)))))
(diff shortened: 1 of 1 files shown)


sionescu:
The semantics of `with-foreign-pointer` is the same as in C: one is not supposed to mutate a pointer returned by malloc and I don't think it's a good idea to hide how things work.

AeroNotix:
C doesn't have 'with-foreign-pointer', so I am unclear to which feature of C you are comparing with, specifically. If `with-foreign-pointer` is supposed to be malloc, then perhaps documentation should be added which makes it clear what the limitations on the provided pointer are.

Still, in my eyes, cffi is in the unique position of being an extremely high level environment calling into lower-level code - provisions could be made which make it much safer to do so.

sionescu:
Yes, `with-foreign-pointer` is supposed to be sugar for malloc+free, just as `with-open-file` is for open+close.
You can't make CFFI much more high-level without runtime overhead because any copy of a pointer still can't be passed to free(). To make that truly safe and nice to use you need reference counting.
CFFI should be nothing more than C-in-Lisp and any other high-level library can build on top of it.

PR #317, opened 2022-03-16
Load only libraries whose architecture matches to architecture of the current lisp process.

svetlyak40wt:
This fixes issue when you are having a CFFI library definition which points
to a few possible library paths, like this one:

```lisp
(cffi:define-foreign-library libcrypto
  ((:and :darwin :cl+ssl-homebrew-found)
   "/usr/local/opt/openssl/lib/libcrypto.dylib")
  ((:and :darwin :cl+ssl-homebrew-arm64-found)
   "/opt/homebrew/opt/openssl/lib/libcrypto.dylib"))
```

On OSX you can have two Homebrew installations with both of these libraries.
Library `/usr/local/opt/openssl/lib/libcrypto.dylib` will have `x86_64` architecture,
whereas `/opt/homebrew/opt/openssl/lib/libcrypto.dylib` will have `arm64`.

Before this patch, if you tried to load the library from the lisp running
in `arm64` more, you've got an error saying that it is impossible to load
`/usr/local/opt/openssl/lib/libcrypto.dylib` library because it does not support `arm64`.

After this patch, CFFI will only try to load a library if it's architecture and
architecture of the current process are the same.

### DIFF
--- a/src/libraries.lisp
+++ b/src/libraries.lisp
@@ -170,11 +170,58 @@
 (defun foreign-library-pathname (lib)
   (slot-value (get-foreign-library lib) 'pathname))
 
+(defun file-architectures (pathname)
+  (when (probe-file pathname)
+    (let* ((output
+             (with-output-to-string (s)
+               (uiop:run-program (list "file" pathname)
+                                 :output s)))
+           (lines (uiop:split-string output :separator '(#\Newline))))
+      (loop for line in lines
+            when (and (> (length line)
+                         0)
+                      (not (char= (elt line (1- (length line)))
+                                  #\])))
+              collect (car (last (uiop:split-string line)))))))
+
+(defun arch-is-compatible (pathname)
+  (loop with current-arch = (machine-type)
+        with file-supports-current-arch = nil
+        with supported = (file-architectures pathname)
+        for arch in supported
+        when (string-equal arch current-arch)
+          do (setf file-supports-current-arch t)
+        finally (return file-supports-current-arch)))
+
+(defun filter-compatible-libs (spec)
+  (typecase spec
+    ((or pathname string)
+     (when (arch-is-compatible spec)
+       spec))
+    (list
+     (case (car spec)
+       (:or (let ((compatible-files
+                    (remove nil
+                            (mapcar #'filter-compatible-libs
+                                    (cdr spec)))))
+              (when compatible-files
+                (list* :or compatible-files))))
+       ;; For other types of spec
+       ;; arch compatibility is not supported yet,
+       ;; thus we'll return them as is
+       (t
+        spec)))))
+
 (defun %foreign-library-spec (lib)
-  (assoc-if (lambda (feature)
-              (or (eq feature t)
-                  (featurep feature)))
-            (slot-value lib 'spec)))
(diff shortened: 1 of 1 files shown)


svetlyak40wt:
Travis build failed with this strange error:

```
 git clone --depth=1 git://github.com/trivial-features/trivial-features.git ~/lisp/trivial-features
Cloning into '/home/travis/lisp/trivial-features'...
fatal: remote error: 
 The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
The command "git clone --depth=1 git://github.com/trivial-features/trivial-features.git ~/lisp/trivial-features" failed and exited with 128 during .
```

Probably CI pipeline was broken.

svetlyak40wt:
Yeah, you are correct about filtering of non existing files. With my patch changes like https://github.com/cl-plus-ssl/cl-plus-ssl/commit/b6eda80a3e79459397085321bfb646f97bc6d848 will become unnecessary.

sionescu:
This is outside the scoope of CFFI because it goes beyond what a C compiler would do.
If you give a C compiler/linker a full path to a library (using -L) it will also complain that the architecture is wrong and not try to auto-detect things.

svetlyak40wt:
This is interesting. How does `dlopen` handle missing files?

svetlyak40wt:
What if instead of this patch we'll made CFFI skip all errors if there are more files in a spec to try? And only if the last file loading failed, signal an error?

daewok:
> New in macOS Big Sur 11.0.1, the system ships with a built-in dynamic linker cache of all system-provided libraries. As part of this change, copies of dynamic libraries are no longer present on the filesystem. Code that attempts to check for dynamic library presence by looking for a file at a path or enumerating a directory will fail. Instead, check for library presence by attempting to dlopen() the path, which will correctly check for the library in the cache. (62986286)

That's from <https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-release-notes>

> This will solve the problem with architecture mismatch, because a library with wrong architecture will be skipped.

Maybe. But it also breaks explicitly documented behavior. Perhaps it's best to add restarts so it's possible to build what you're describing on top of what CFFI provides? But I'm just throwing out ideas with no inkling of if they're likely to be accepted. :)

Also, I don't think this will fix things for cl-plus-ssl. We've already put together that Apple has a bug/feature (you never know with Apple!) such that `dlopen("/some/non-existent/path/to/libcrypto.dylib")` is instead treated the same as `dlopen("libcrypto.dylib")`, which aborts the process pretty hard.

Based on the latest activity in cl-plus-ssl/cl-plus-ssl#138, it seems that `dlopen("/path/to/wrong/arch/libcrypto.dylib")` has that same bug/feature, so CFFI would never get the chance to gracefully handle the error.

svetlyak40wt:
Ok. I'll experiment with that. Probably the best way would be to run dlopen in a subprocess first just to check if it is able to load library. This way the main process will not hang or crash.

svetlyak40wt:
@sionescu @daewok I've implemented another approach and it works for me on Apple M1 and two Homebrew installation having openssl for x86 and ARM:

https://github.com/cffi/cffi/pull/318

Please, review that pull.

PR #318, opened 2022-03-19
Try to load all specs, matching to the current *features*.

svetlyak40wt:
This should fix a problem of having libraries for different architectures
on Apple M1.

### DIFF
--- a/src/libraries.lisp
+++ b/src/libraries.lisp
@@ -201,8 +201,16 @@
                           (t nil)))))
 
 (defun foreign-library-options (lib)
+  ;; Here we are collecting options from all matched specs.
+  ;; Probably this is not correct because different specs
+  ;; might have different call conventions and search paths.
+  ;; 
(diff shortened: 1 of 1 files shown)


svetlyak40wt:
@sionescu please, review this pull?

sionescu:
As it is, this is not going in.

svetlyak40wt:
Why not?

svetlyak40wt:
@sionescu what was wrong with this pull. What should be fixed?

luismbo:
@svetlyak40wt `define-foreign-library` was designed and is documented to work like a `cond` and existing code depends on that. (e.g., they may have specific declarations for openbsd and a generic one for other unix systems).

Perhaps you'd like to start a [discussion](https://github.com/cffi/cffi/discussions) about the Apple M1 issue you're having and what CFFI can do to help, if anything.

PR #336, opened 2022-09-05
Fix / Improve with-foreign-pointer-as-string in manual

### LINKED ISSUE #335: Update documentation for with-foreign-pointer-as-string
`with-foreign-pointer-as-string` has incorrect documentation in the manual. In order to place the length of the buffer into a variable, it suggest the following syntax:

``` lisp
(with-foreign-pointer-as-string (str 6 str-size :encoding :ascii)
     (lisp-string-to-foreign "Hello, foreign world!" str str-size)))
```

However, it should actually look like this:

``` lisp
(with-foreign-pointer-as-string ((str str-size) 6 :encoding :ascii)
     (lisp-string-to-foreign "Hello, foreign world!" str str-size)))
```
The syntax needs to be corrected too.

Here's the current macro definition for reference: https://github.com/cffi/cffi/blob/e0b2f359666465bcd49432b9e596b38b9089c9a7/src/strings.lisp#L228

sdilts:
+ Fix syntax and example to show the correct syntax
+ Add missing macro parameter descriptions
+ Rewrite the macro description to be a bit more clear on exactly what
  the macro is doing.

Fixes #335.

### DIFF
--- a/doc/cffi-manual.texinfo
+++ b/doc/cffi-manual.texinfo
@@ -5082,39 +5082,55 @@ CFFI> (with-foreign-string (foo "12345")
 @node with-foreign-pointer-as-string,  , with-foreign-string, Strings
 @heading with-foreign-pointer-as-string
 @subheading Syntax
-@Macro{with-foreign-pointer-as-string (var size &optional size-var @
+@Macro{with-foreign-pointer-as-string (var-or-vars size @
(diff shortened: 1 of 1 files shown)


sdilts:
Is there something blocking this from getting merged?

luismbo:
Sorry for the delay. Thanks!

PR #343, opened 2022-11-07
Add compiler macro foreign-type-size

### LINKED ISSUE #340: Expanding foreign-type-size for standard types
On a fresh SBCL instance, the function `foreign-type-size` is called, even for constant expressions eval'd to standard types:

```lisp
* (require :cffi)
NIL
* (defun foo ()
    (declare (optimize speed))
    (cffi:foreign-type-size :int))
FOO
* (disassemble #'foo)
; disassembly for FOO
; Size: 22 bytes. Origin: #x5366769B                          ; FOO
; 9B:       BA5F873F50       MOV EDX, #x503F875F              ; ':INT
; A0:       B902000000       MOV ECX, 2
; A5:       FF7508           PUSH QWORD PTR [RBP+8]
; A8:       B8629C3F50       MOV EAX, #x503F9C62              ; #<FDEFN CFFI:FOREIGN-TYPE-SIZE>
; AD:       FFE0             JMP RAX
; AF:       CC10             INT3 16                          ; Invalid argument count trap
```

While a custom type may be redefined to be a different size, I don't think there ought to be a reason why types like `:int`, `:uint` etc should change, especially for implementations whose compiled output aren't portable across architectures.

Here's a first pass what I think a compiler macro for this could look like:

```lisp
(define-compiler-macro foreign-type-size (&whole whole foreign-type
                                                 &environment env)
    (flet ((standard-type-p (type)
             (or (keywordp type)
                 (and (lisp type) (eq (first type) :pointer)))))
      (let ((expanded-type (macroexpand foreign-type env))
            evaled-type)
        (if (and (constantp expanded-type env)
                 (standard-type-p (setf evaled-type (eval expanded-type)))
                 (ignore-errors (foreign-type-size evaled-type)))
            (foreign-type-size evaled-type)
            whole))))
```

samuel-hunter:
Fixes #340. Depends on #345 

The compiler macro is a first-pass that should fix most use cases without breakage. It checks to see if the foreign type is constant, and if so, expands into its foreign type size, falling back to the old form if an error is signaled.

Before the compiler macro:

```lisp
* (disassemble '(lambda () (cffi:foreign-type-size '(:pointer :uint32))))
; disassembly for (LAMBDA ())
; Size: 32 bytes. Origin: #x5365FC5C                          ; (LAMBDA ())
; 5C:       498B4510         MOV RAX, [R13+16]                ; thread.binding-stack-pointer
; 60:       488945F8         MOV [RBP-8], RAX
; 64:       488B15BDFFFFFF   MOV RDX, [RIP-67]                ; '(:POINTER
                                                              ;   :UINT32)
; 6B:       B902000000       MOV ECX, 2
; 70:       FF7508           PUSH QWORD PTR [RBP+8]
; 73:       B8A29D3F50       MOV EAX, #x503F9DA2              ; #<FDEFN CFFI:FOREIGN-TYPE-SIZE>
; 78:       FFE0             JMP RAX
; 7A:       CC10             INT3 16                          ; Invalid argument count trap
NIL
```

After (on SBCL 2.2.5 64-bit)

```lisp
* (disassemble (lambda () (cffi:foreign-type-size '(:pointer :uint32))))
; disassembly for (LAMBDA ())
; Size: 21 bytes. Origin: #x5365FECC                          ; (LAMBDA ())
; CC:       498B4510         MOV RAX, [R13+16]                ; thread.binding-stack-pointer
; D0:       488945F8         MOV [RBP-8], RAX
; D4:       BA10000000       MOV EDX, 16
; D9:       488BE5           MOV RSP, RBP
; DC:       F8               CLC
; DD:       5D               POP RBP
; DE:       C3               RET
; DF:       CC10             INT3 16                          ; Invalid argument count trap
NIL
```

Tested on SBCL and CCL with no additional test errors.

### DIFF
--- a/src/early-types.lisp
+++ b/src/early-types.lisp
@@ -164,6 +164,11 @@ Signals an error if FOREIGN-TYPE is undefined."))
   (:documentation
    "Return the size in bytes of a foreign type."))
 
+(define-compiler-macro foreign-type-size (&whole form foreign-type)
+  (if (constant-form-p foreign-type)
+      (foreign-type-size (constant-form-value foreign-type))
+      form))
+
 (defgeneric unparse-type (foreign-type)
   (:documentation
    "Unparse FOREIGN-TYPE to a type specification (symbol or list)."))
--- a/src/early-types.lisp
+++ b/src/early-types.lisp
@@ -629,24 +629,24 @@ Signals an error if the type cannot be resolved."
   (translate-to-foreign value (parse-type type)))
 
 (define-compiler-macro convert-to-foreign (value type)
-  (if (constantp type)
-      (expand-to-foreign value (parse-type (eval type)))
+  (if (constant-form-p type)
+      (expand-to-foreign value (parse-type (constant-form-value type)))
       `(translate-to-foreign ,value (parse-type ,type))))
 
 (defun convert-from-foreign (value type)
   (translate-from-foreign value (parse-type type)))
 
 (define-compiler-macro convert-from-foreign (value type)
-  (if (constantp type)
-      (expand-from-foreign value (parse-type (eval type)))
+  (if (constant-form-p type)
+      (expand-from-foreign value (parse-type (constant-form-value type)))
       `(translate-from-foreign ,value (parse-type ,type))))
 
 (defun convert-into-foreign-memory (value type ptr)
   (translate-into-foreign-memory value (parse-type type) ptr))
 
 (define-compiler-macro convert-into-foreign-memory (value type ptr)
-  (if (constantp type)
-      (expand-into-foreign-memory value (parse-type (eval type)) ptr)
+  (if (constant-form-p type)
+      (expand-into-foreign-memory value (parse-type (constant-form-value type)) ptr)
       `(translate-into-foreign-memory ,value (parse-type ,type) ,ptr)))
 
 (defun free-converted-object (value type param)
--- a/src/enum.lisp
+++ b/src/enum.lisp
@@ -321,8 +321,8 @@
 (define-compiler-macro foreign-bitfield-value (&whole form type symbols)
   "Optimize for when TYPE and SYMBOLS are constant."
   (declare (notinline foreign-bitfield-value))
-  (if (and (constantp type) (constantp symbols))
-      (foreign-bitfield-value (eval type) (eval symbols))
+  (if (and (constant-form-p type) (constant-form-p symbols))
+      (foreign-bitfield-value (constant-form-value type) (constant-form-value symbols))
       form))
 
 (defun %foreign-bitfield-symbols (type value)
@@ -348,8 +348,8 @@ the bitfield TYPE."
 (define-compiler-macro foreign-bitfield-symbols (&whole form type value)
   "Optimize for when TYPE and SYMBOLS are constant."
   (declare (notinline foreign-bitfield-symbols))
-  (if (and (constantp type) (constantp value))
-      `(quote ,(foreign-bitfield-symbols (eval type) (eval value)))
+  (if (and (constant-form-p type) (constant-form-p value))
+      `(quote ,(foreign-bitfield-symbols (constant-form-value type) (constant-form-value value)))
       form))
 
 (defmethod translate-to-foreign (value (type foreign-bitfield))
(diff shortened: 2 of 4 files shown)


src/early-types.lisp
```diff
@@ -164,6 +164,14 @@ Signals an error if FOREIGN-TYPE is undefined."))
   (:documentation
    "Return the size in bytes of a foreign type."))
 
+(define-compiler-macro foreign-type-size (&whole form foreign-type)
+  ;; Don't pass &environment to CONSTANTP or MACROEXPAND as it may make
+  ;; surprising behavior with LOAD-TIME-VALUE.
+  (if-let ((comptime-value (when (constantp (macroexpand foreign-type))
+                             (ignore-errors (eval form)))))
```
phoe, line 171: `ignore-errors` + `eval` looks like a pretty big code smell to me; wouldn't `(locally (declare (notinline foreign-type-size)) (foreign-type-size foreign-type))` work better? It should ensure that the function is called at compilation time without recursively invoking the compiler macro again.

src/early-types.lisp
```diff
@@ -164,6 +164,14 @@ Signals an error if FOREIGN-TYPE is undefined."))
   (:documentation
    "Return the size in bytes of a foreign type."))
 
+(define-compiler-macro foreign-type-size (&whole form foreign-type)
+  ;; Don't pass &environment to CONSTANTP or MACROEXPAND as it may make
+  ;; surprising behavior with LOAD-TIME-VALUE.
```
phoe, line 169: Could you provide more examples of this?

src/early-types.lisp
```diff
@@ -164,6 +164,14 @@ Signals an error if FOREIGN-TYPE is undefined."))
   (:documentation
    "Return the size in bytes of a foreign type."))
 
+(define-compiler-macro foreign-type-size (&whole form foreign-type)
```
phoe, line 167: Are there any tests that should be performed for this compiler macro alone? I assume that you could call the compiler macro function yourself to ensure that it provides correct results on some test cases.

src/early-types.lisp
samuel-hunter: Oh, nice! This was what I was looking for before I gave up on `(l-t-v ,form)` -> eval.

src/early-types.lisp
samuel-hunter: I rephrased this now that I took a second look at this. Clarified: since L-T-V executes under a null lexical environment, we shouldn't be checking if the foreign type is constant under &environment.

sionescu:
@samuel-hunter Copy QUOTEDP, CONSTANTP, and EVAL-CONSTANT from https://github.com/sionescu/static-vectors/blob/master/src/constantp.lisp and use them.
We'll figure out later how to share the code.

samuel-hunter:
> @samuel-hunter Copy QUOTEDP, CONSTANTP, and EVAL-CONSTANT from https://github.com/sionescu/static-vectors/blob/master/src/constantp.lisp and use them. We'll figure out later how to share the code.

Preexisting compiler macros ought to benefit from this if grafted into CFFI. Should this be split into its own PR?

sionescu:
Why a new PR ?

samuel-hunter:
I'll make a new branch/PR at least for now

sionescu:
@samuel-hunter Please rebase to master. I added the utils in d4216a3acafa.

sionescu:
I opened https://github.com/cffi/cffi/issues/344 for converting existing compiler macros.

src/early-types.lisp
```diff
@@ -164,6 +164,13 @@ Signals an error if FOREIGN-TYPE is undefined."))
...
+  ;; Don't pass &environment to CONSTANTP or MACROEXPAND as LOAD-TIME-VALUE is
+  ;; executed in a null lexical environment.
+  (if (constantp (macroexpand foreign-type))
+      `(load-time-value (locally (declare (notinline foreign-type-size)) ,form))
```
sionescu, line 171: It's better to use CONSTANT-FORM-VALUE here. L-T-V inhibits futher optimisations, such as (foreign-alloc (+ <constant> (foreign-type-size :foo)))

samuel-hunter: Fixed, thanks :+1:

src/early-types.lisp
```diff
@@ -165,10 +165,8 @@ Signals an error if FOREIGN-TYPE is undefined."))
    "Return the size in bytes of a foreign type."))
 
 (define-compiler-macro foreign-type-size (&whole form foreign-type)
-  ;; Don't pass &environment to CONSTANTP or MACROEXPAND as LOAD-TIME-VALUE is
-  ;; executed in a null lexical environment.
-  (if (constantp (macroexpand foreign-type))
-      `(load-time-value (locally (declare (notinline foreign-type-size)) ,form))
+  (if (constant-form-p (macroexpand foreign-type))
```
sionescu, line 168: Let's do the conversion in another PR. You can cherry-pick this commit to another branch.

sionescu (changes requested):

src/early-types.lisp
sionescu: Did you undo the change to use CONSTANT-FORM-VALUE ?

sionescu (changes requested):

src/early-types.lisp
samuel-hunter: I don't *think* so...? Unless I did that in process of reordering the commits.

PR #345, opened 2022-11-08
Switch compiler macros to CONSTANT-FORM-VALUE instead of CL:EVAL

samuel-hunter:


samuel-hunter:
All uses of CL:EVAL are switched out except for the CFFI-SYS macros in `cffi-*.lisp`. Two ideas I've had for moving CONSTANT-FORM-VALUE to let these macros use it as well are:

- Create a package cffi-sys-util that both cffi-sys and cffi can import from (would exposing a new package be problematic?)
- For each `cffi-*.lisp`, split them into head and tail files and have CONSTANT-FORM-VALUE loaded in-between (Prevents exposure of a new package, but doubling files here *feels* like a code smell here)

I'm leaning on option 1. Any thoughts?

sionescu:
8629618e9bb0 refactors the package definitions and exports CONSTANT-FORM-P and CONSTANT-FORM-VALUE from CFFI-SYS.

samuel-hunter:
Switched out functions in cffi-sys. This should be ready for another round of reviews

PR #356, opened 2023-03-20
Add additional DLL names to libffi Windows spec

kartik-s:
I want to use cffi in a Conda environment on Windows, but the Conda-distributed package for libffi drops the `lib` prefix when installing the DLL:

```
>>> Library/bin/ffi-8.dll <<<
Library/include/ffi.h
Library/include/ffitarget.h
Library/lib/ffi.lib
Library/lib/libffi.lib
Library/lib/pkgconfig/libffi.pc
Library/share/info/libffi.info
Library/share/man/man3/ffi.3
Library/share/man/man3/ffi_call.3
Library/share/man/man3/ffi_prep_cif.3
Library/share/man/man3/ffi_prep_cif_var.3
```

This PR makes it so that cffi can find these DLLs.

### DIFF
--- a/libffi/libffi.lisp
+++ b/libffi/libffi.lisp
@@ -32,7 +32,8 @@
   (:solaris (:or "/usr/lib/amd64/libffi.so" "/usr/lib/libffi.so"))
   ((:or :netbsd :freebsd :openbsd) "libffi.so")
   (:unix (:or "libffi.so.8" "libffi32.so.8" "libffi.so.7" "libffi32.so.7" "libffi.so.6" "libffi32.so.6" "libffi.so.5" "libffi32.so.5" "libffi.so" "libffi32.so"))
-  (:windows (:or "libffi-8.dll" "libffi-7.dll" "libffi-6.dll" "libffi-5.dll" "libffi.dll"))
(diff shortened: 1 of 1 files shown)


luismbo:
Did you consider other alternatives such as (a) including a copy of libffi next to your Lisp's binary or (b) symlinking ffi-8.dll to libffi-8.dll?

macrologist:
> Did you consider other alternatives such as (a) including a copy of libffi next to your Lisp's binary or (b) symlinking ffi-8.dll to libffi-8.dll?

Why are these alternatives considered to be superior? What is the harm in this PR?

luismbo:
Not necessarily superior, but it's a bit weird that we have to keep track of some arbitrary renaming by the Conda project. If one of the alternatives would be acceptable, that'd be simpler CFFI. But, no, it's not particularly harmful to list those alternatives, I suppose. Maybe a comment is in order to clarify why those alternatives are listed. Also, Conda seems recent enough to not ever have shipped all of those earlier versions of libffi. Maybe adding ffi-8.dll is enough.

macrologist:
Fair enough.

I'm pretty sure that ffi-7.dll shipped in the version 3.3 of the libffi available through conda forge. ffi-8.dll ships in version 3.4.2.

@kartik-s

digikar99:
Bumping this, it'd be nice for cffi to take care of this! I don't see any downsides.

PR #357, opened 2023-04-17
Support loading cffi as monolithic compile-bundle into a clean Lisp

gendl:
Protect against loading this as a monolithic compile bundle which doesn't have asdf in it.


luismbo:
Does that fail when the fasl references an inexistent ASDF package? Two suggestions: drop the commented out code and maybe use uiop:symbol-call?

gendl:
Hi Luis,

The original version was causing builds to fail if the build is using a compile bundle (asdf's monolithic-compile-bundle-op) for cffi, but where asdf itself had not been loaded into the build, e.g. for runtime builds where we don't need the asdf or quicklisp tooling itself included in the build. 

I deleted the commented-out previous line and my comment, added a check for existence of `asdf` package, and changed the `(funcall (read-from-string ...))`` into `uiop:symbol-call` as directed.  

Please have a look at it now and let me know your thoughts,

Thanks,

 Dave

PR #374, opened 2024-01-02
try to translate constant enum keywords at compile time

3b:


luismbo:
It'd be nice to use a constantp-aware once-only, but not a big deal. CI seems to be MIA. Did you run the test suite?

sionescu:
@luismbo our free Travis credits have run out and we should probably migrate to Github actions.

3b:
hadn't run tests, was hoping CI would take care of it so I don't have to deal with windows/msys oddness.

Trying now says

```
Unhandled memory fault at #x500000020.
   [Condition of type CFFI-GROVEL:GROVEL-ERROR]
```
from trying to build libffi (and possibly also from running pkg-conf, libffi is installed and pkg-conf seems to be working).

Will try to debug that and/or run tests elsewhere later.

luismbo:
Feel free to skip the libffi tests.

3b:
It looked more like a problem with `run-program` than libffi specifically, but will try that when i get back to the computer.

3b:
Yeah, something is wrong with my local setup, i get segfaults from run-program half the time. When tests do run, i get 12 unexpected failures with or without the changes, some look like the run-program problem, some from sprintf failing (`%0.f` always prints "f" or "0"), and a few other random things.

3b:
on x8664 linux sbcl i get
```
3 out of 342 total tests failed: FSBV.WFO, FSBV.MAKEPAIR.1, FSBV.MAKEPAIR.2.
No unexpected failures.
```
with this branch

PR #391, opened 2024-07-12
make sure structs are big enough to hold all slots even if :offset is used

3b:


tests/struct.lisp
```diff
@@ -713,3 +713,29 @@
...
+  (f :uint8 :offset 6)
+  (g :uint32 :offset 0))
+
+(deftest struct.offset.1
+    (foreign-type-size '(:struct struct.offsets))
+  20)
+
+;; Add some tests for :offset in general
+(deftest struct.offset.2
+    (values (cffi:foreign-slot-offset '(:struct struct.offsets) 'a)
```
luismbo, line 734: No need for the package prefix, I guess. We don't use it elsewhere.

luismbo (changes requested):
Nice edge case.

tests/struct.lisp
```diff
@@ -713,3 +713,29 @@
...
+            (cffi:foreign-slot-offset '(:struct struct.offsets) 'b)
+            (cffi:foreign-slot-offset '(:struct struct.offsets) 'c)
+            (cffi:foreign-slot-offset '(:struct struct.offsets) 'd)
+            (cffi:foreign-slot-offset '(:struct struct.offsets) 'e)
+            (cffi:foreign-slot-offset '(:struct struct.offsets) 'f)
+            (cffi:foreign-slot-offset '(:struct struct.offsets) 'g))
```
luismbo, line 740: Maybe use mapcar across each slot name? 🤷‍♂️

tests/struct.lisp
3b: oops, added it to get autodoc since i hadn't loaded the package, and forgot to remove it

tests/struct.lisp
3b: done

3b:
added a few more slots to test some other properties of `:offset` as well

luismbo (approved):

PR #392, opened 2024-07-16
Add support for ECL long double

yitzchak:
Also fix feature expression for SCL that used long-double instead of long-float.


luismbo:
Thanks!

Issue #381, opened 2024-04-06
Compatibility with Allegro v.11


RealSharpness:
I use Common Lisp library lispbuilder-sdl which relies on CFFI.

My lisp implementation Allegro CL has updated to version 11 and their internal symbol FF-FUNCALL has changed to FF_FUNCALL. (Not certain, but I think this is because it can now be used in the browser.)

For reference, here is the lispbuilder-sdl code which calls FF-FUNCALL.
https://github.com/cffi/cffi/blob/621cb9d5d6f49f91d2d6ff05044ea80e3ea6b67f/src/cffi-allegro.lisp#L260

I am a bit out of my depth in this topic, but would it be possible to update CFFI so the new symbol FF_FUNCALL is used?

phoe:
Sure, we'll need someone who can write a fixed implementation conditionalized on ACL version, and then then run all the tests for both old and new ACL. (Won't be me, since I'm not an ACL user; does Franz have any sort of user forum where this issue might have already been raised or where we can ask for a list of FFI-related changes that would speed things up on CFFI side?)

RealSharpness:
Thank you,

I have not found a user forum and believe Franz handles all matters directly with individual customers.

https://franz.com/support/documentation/11.0/foreign-functions.html#30-the-foreign-function-interface-specification
This section of v.11 documentation describes ACL's FFI which is mostly the same on all versions "except as noted." The changed symbol which produced this error however is an undocumented internal symbol.

I can try requesting anything useful that isn't covered in the documentation, please let me know. I'll provide whatever support I can.

phoe:
Sure, please try linking them to this issue.

RealSharpness:
I sent a follow-up e-mail to this issue that Franz helped me identify but haven't received a response yet.

I'm wondering if there's any ACL user here who can help me with this problem. I suppose there isn't any great demand at this time but I would offer what compensation I can.

LEHunter:
Yes, I am an ACL user who just ran into the same problem.  I hadn't yet figured out the new symbol though, so thanks.  I don't have an old version to run the tests on, but I can write a fix conditionalized on ACL version, and then then run all the tests for  new ACL.

LEHunter:
Well, I can help you, but not do the heavy lifting for the update. Change these two functions in src/cffi-allegro.lisp

```(defmacro %foreign-funcall (name args &key convention library)
  (declare (ignore convention library))
  (multiple-value-bind (types fargs rettype)
[...]

PR #398, opened 2024-10-28
Initial fix for Allegro CL 11

dieggsy:
The internal system::ff_funcall function was renamed to system::ff_funcall. This gets us most of the way there, but there are still issues with vararg functions.

There are still some outstanding issues ([tests that fail in 11 but not in 10](https://github.com/cffi/cffi/issues/381#issuecomment-2181397061)), but this seems better than nothing for now.

### DIFF
--- a/src/cffi-allegro.lisp
+++ b/src/cffi-allegro.lisp
@@ -257,7 +257,8 @@ WITH-POINTER-TO-VECTOR-DATA."
   (declare (ignore convention library))
   (multiple-value-bind (types fargs rettype)
       (foreign-funcall-type-and-args args)
-    `(system::ff-funcall
+    `(#+(version>= 11 0) system::ff_funcall
+      #-(version>= 11 0) system::ff-funcall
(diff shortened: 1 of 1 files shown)


dieggsy:
I should note that while this doesn't 100% fix `foreign-funcall`, it does make CFFI usable at all on ACL 11. I do think further fixes can be made separately.

dieggsy:
Actually, CFFI does seem to still work on ACL11 as long as `%foreign-funcall` is not called. For example the CFFI ODBC port at [sharplispers/clsql](https://github.com/sharplispers/clsql) is loading and working ok.

@luismbo @sionescu ideally I'd still like feedback on how to proceed with this though.

luismbo:
Can you simplify the commit message to `cffi-allegro: use SYSTEM::FF_FUNCALL for ACL >= 11`? 

(if I do it through GitHub it'll claim authorship)

dieggsy:
@luismbo done

luismbo:
Thanks. Sorry for the delay.

PR #400, opened 2024-11-03
cffi doesn't provide correct include switches to compiler under FreeBSD, NetBSD and OpenBSD.

tapioco71:


sionescu:
@tapioco71 Please make this a single commit with a nice message.

PR #401, opened 2024-11-14
grovel: don't treat pkg-config stderr as compilation flags

lockie:
Hey. I'm being bitten by the fact that grovel subsystem interprets `pkg-config` stderr as the compilation flags for GCC, while its stderr usually contains some kind of warnings. 

This small PR fixes the problem by redirecting stderr of `pkg-config` called by groveller to `*debug-io*`.


luismbo:
Thanks. Sorry for the delay.

lockie:
No problem, thanks!

PR #404, opened 2025-01-28
Avoid dependency on UIOP in CFFI core

Shinmera:
The first patch is unrelated, but still handy since I noticed that in some libraries repeated parse-type use can be quite allocation-happy. If you think it's a waste, I can remove it again.

The second two patches remove the dependency on UIOP from the core ASDF system by factoring out the darwin frameworks related functions that make use of it into a separate file and featuring it away.

UIOP and ASDF both can be quite a headache for deployments and it would give me a lot of peace of mind to be able to drop dependency on it. With some extra work the dependency could be avoided on darwin as well, but I'd rather not copy UIOP functionality into CFFI unless it is desired to sever the dependency more aggressively. If so, let me know, and I can do the work.

In either case I think the change to factor out darwin framework handling to a separate file should be fine, since it makes no sense on any platform other than darwin.


luismbo:
> The first patch is unrelated, but still handy since I noticed that in some libraries repeated parse-type use can be quite allocation-happy. If you think it's a waste, I can remove it again.

I think it's fair to expect CFFI to be as efficient as possible, although `parse-type` at runtime is unexpected in that context. The change is fine.


> The second two patches remove the dependency on UIOP from the core ASDF system by factoring out the darwin frameworks related functions that make use of it into a separate file and featuring it away.
> 
> UIOP and ASDF both can be quite a headache for deployments and it would give me a lot of peace of mind to be able to drop dependency on it. With some extra work the dependency could be avoided on darwin as well, but I'd rather not copy UIOP functionality into CFFI unless it is desired to sever the dependency more aggressively. If so, let me know, and I can do the work.

If UIOP is that painful perhaps it's worth leaving a comment in the cffi defsystem that we're purposefully avoiding it. I'm fine with leaving it in for darwin until it bothers someone else.


> In either case I think the change to factor out darwin framework handling to a separate file should be fine, since it makes no sense on any platform other than darwin.

Agreed! That worked out nicely.

PR #412, opened 2025-08-19
fix for cmucl: close the correct library

vibs29:
cffi's library handle is set to whatever %load-foreign-library returns. That same handle is passed to %close-foreign-library.

%load-foreign-library returns an index into cmucl's *global-table*. %close-foreign-library wants to index into the *global-table*, for which it can and should use the handle directly rather than trying to change it via unix-namestring.

In particular, for a library loaded from /lib or the LD_LIBRARY_PATH (rather than in the current directory or cffi:*foreign-library-directories*), the handle (i.e. index into *global-table*) is _not_ a relative or absolute pathname and thus applying unix-namestring to it produces NIL. That prevents the correct *global-table* entry from being closed.


luismbo:
Thanks.

PR #413, opened 2025-08-20
Look up CallbackReference class robustly

vibs29:
JSS has a table of class names that it builds by scanning jar files on the classpath but not directories. (Maybe that's on purpose in case the user has on the classpath a large directory intended to be probed for class files but not recursively scanned.)

So rather than rely on such heuristics from JSS, it's more robust to pass JSS the full correct case-sensitive class name. That lets cffi work even when the JNA library is provided as a directory of class files instead of a jar file. Otherwise cffi only works when JNA is provided as a jar file.


luismbo:
Thanks

PR #420, opened 2025-11-24
Speed up WITH-POINTER-TO-VECTOR-DATA

vibs29:
Making a foreign call (%MEM-REF/%MEM-SET) per element is slow. Instead, by using MEMORY-AS to copy the whole array in one foreign call, the macro called with an empty body runs 3.5x as fast with a 1 KiB buffer and 33.9x as fast with a 1 MiB buffer.


luismbo:
Thanks @vibs29

PR #430, opened 2026-05-16
Simplify single-bit-p

stassats:
Performance doesn't matter here, but the current definition is quite bad.


luismbo:
Yes indeed. 😅

PR #435, opened 2026-06-24
Update defcenum documentation

sdilts:
+ Mention that `defcenum` produces constant definitions when non-keyword symbols are used for the enum value identifiers.
+ Update section on the underlying base type; CFFI follows the C standard, which doesn't guarantee that the underlying type is an `int`.
  + Also note that Windows doesn't always follow the standard.


luismbo:
Thanks!

